PR #20, opened 2026-05-18
Infer repo_source and pipeline from task.toml when pushing to HF Hub

sergiopaniego:
## Problem

When running `repo2rlenv push`, the source repo and pipeline were not passed to `push_to_hub`, so the generated README and registry.json had empty/placeholder values:

- README: `**Source repo**: [``](https://github.com/)`
- registry.json: `"description": "Repo2RLEnv pr_diff from "`

Example: https://huggingface.co/datasets/sergiopaniego/trl-r2e-test

## Fix

`cmd_push` now reads the first `task.toml` it finds in the dataset directory and extracts `metadata.repo2env.repo` and `metadata.repo2env.pipeline` before calling `push_to_hub`. Both fields are already present in every task generated by Repo2RLEnv pipelines.

The fallback behavior (empty repo_source, pipeline="pr_diff") is unchanged, so existing calls are not affected.

## Testing

Generated a dataset from `huggingface/trl` via `pr_diff`, pushed to HF Hub, and verified the README and registry.json are correctly populated.

### DIFF
--- a/src/repo2rlenv/cli.py
+++ b/src/repo2rlenv/cli.py
@@ -459,6 +459,22 @@ def cmd_push(args: argparse.Namespace) -> int:
             f"--push does not pin revisions; ignoring `@{revision}` (Hub commit will be the push commit)"
         )
 
+    # Infer repo_source and pipeline from the first task.toml found in the dataset dir
+    repo_source = ""
+    pipeline = "pr_diff"
+    for task_toml in sorted(local_dir.rglob("task.toml")):
+        try:
+            import tomllib
+
+            with open(task_toml, "rb") as f:
+                meta = tomllib.load(f)
+            r2e = meta.get("metadata", {}).get("repo2env", {})
+            repo_source = r2e.get("repo", "")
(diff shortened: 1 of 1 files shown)


adithya-s-k:
Hey @sergiopaniego — thank you for catching this! 🙏

You and I spotted the same bug independently. I took inspiration from your fix and rolled it into #28 (the v0.8.2.post3 image-distribution PR), which also threads in registry-backed reproducibility:

- `push_to_hub` now auto-reads `pipeline` + `repo` from the first task's `[metadata.repo2env]` subtable (caller-supplied kwargs still win for back-compat).
- On top of that, every `_runtime` task now gets a new `[metadata.repo2env.reproducibility]` block recording the bootstrap image digest, visibility, and inline-recipe fallback metadata so consumers know exactly what they're getting.
- 7 new tests under `tests/test_hub_metadata_read.py` cover the auto-read + caller-override cases.

Let me know if this is good to close — and again, thank you for the find!

sergiopaniego:
sure, we can close it! thanks!!

PR #22, opened 2026-05-18
clarify harbor run flags and pr_diff limitation

sergiopaniego:
Fix two errors in the harbor run example: wrong -d/-e/-a short flags (harbor uses --path/--env/--agent), and -e local-docker is not a valid environment name (correct: --env docker).

Add a note in both README and quickstart that harbor run requires a sandbox-verified pipeline — pr_diff tasks have no environment/ dir and cannot be executed by Harbor.


adithya-s-k:
thanks @sergiopaniego for correcting the Harbor CLI examples!

PR #23, opened 2026-05-18
fix: bootstrap saves executable test_cmds; normalize strips -q

sergiopaniego:
Bootstrap prompt previously instructed the agent to save `--collect-only` as test_cmds. Downstream pipelines (mutation_bugs,  pr_runtime) need actual test execution with per-test pass/fail output to find baseline passing tests and F2P signal — collect-only produces neither.

Two related fixes in normalize_test_cmds_for_runtime: 
- Already stripped --collect-only; now also strips -q/--quiet, which cancels -v in pytest 9's verbosity counter, causing the log parser to see 0 passing tests even when tests ran.

Existing cached bootstraps are safe: normalize handles both flags so no forced rebuild is required.


adithya-s-k:
thanks @sergiopaniego for fixing the bootstrap test commands and quiet-flag handling, appreciate it!

PR #24, opened 2026-05-18
fix: omit seed from task.toml when not set

sergiopaniego:
seed is int | None (defaults to None for time-based RNG). Writing None directly to the TOML payload crashes tomli_w with TypeError. Omit the key entirely when seed is not set, consistent with the pattern used for synthesis_llm across other pipelines.


adithya-s-k:
thanks @sergiopaniego for catching the unset seed issue!

PR #48, opened 2026-05-28
docs: add reward signal column to pipeline table + Harbor Visualizer

sergiopaniego:


adithya-s-k:
thanks @sergiopaniego for making the reward signals clearer and linking the Harbor visualizer!

PR #49, opened 2026-05-28
fix(commit_runtime): warn when candidate count reaches clone_depth

sergiopaniego:
~~commit_runtime was only stripping "Closes #N" trailers, leaving
cherry-pick refs, backport URLs, and bare SHAs in the instruction
text. Apply the same _strip_info_leak() already used by pr_runtime. (not sure this was intentional)~~

Also warn when candidate count reaches clone_depth, which indicates
the shallow clone may be truncating history.


adithya-s-k:
thanks @sergiopaniego for adding the clone-depth warning, helpful when mining larger histories!

PR #51, opened 2026-05-28
docs: link Verifiable RL Environments HF collection in README

sergiopaniego:


adithya-s-k:
thanks @sergiopaniego for adding the collection link, makes the example environments easier to find!

PR #85, opened 2026-07-23
feat(pipelines): add pr_to_env + v2 iptables egress firewall

### LINKED ISSUE #84: Add pr_to_env pipeline: curated PR URL → Harbor RL env (with v2 iptables egress firewall)
## Motivation

Repo2RLEnv currently mines PR history via `pr_runtime` and applies filters to pick tasks. For **dataset curation** — where you want a hand-picked set of PRs with deliberate difficulty/domain coverage — mine-then-filter is the wrong tool.

Building [HF_ML_Bench_v0](https://huggingface.co/datasets/AdithyaSK/HF_ML_Bench_v0) surfaced this concretely: the audit doc at `plans/HF_PR_to_env.md` picked 114 healthy PRs across `transformers`/`accelerate`/`diffusers`/`peft`/`trl`, and driving those through `pr_runtime` meant fighting the filters and losing the URL→env mapping.

## Proposal

Add `pr_to_env`, the **import-shape sibling** of `pr_runtime`:
- Input: explicit `url=<gh-pr-url>` or `urls_file=<path>`
- Output: one Harbor task per URL, or fail-closed with a per-URL reason
- Same task shape, same graded F2P/P2P verifier, same anti-contamination guards

## What this ships

- `PrToEnvPipeline` (registered as `pipeline="pr_to_env"`, `experimental=True`) at `src/repo2rlenv/pipelines/pr_to_env.py`
- `PrToEnvOptions` with 13 knobs
- **Oracle-gate in-loop**: `harbor run -a oracle` after each emit; sub-1.0 envs dropped + rmtree'd
- `keepers.jsonl` ledger writer
- **Leak-strip v2**: hard-strips bare short-SHAs (8-39 hex) and pytest node-id paths that leak fix targets; soft-flags basename/dirname mentions
- **Dockerfile gate #5 (pyproject sanitize)**: strips bare `[tool.pytest]` when `[tool.pytest.ini_options]` also present (hit repeatedly on peft PRs — pytest exits with usage-error 4 otherwise)
- **Egress firewall v2** in `_env_guard.py`: iptables OUTPUT DROP + IP allow-list (see below)
- 34 new/updated tests (`test_pipeline_pr_to_env.py`, `test_env_guard.py`)

## Anti-contamination: v2 egress firewall

The v1 guard (`egress_guard_compose`) uses `/etc/hosts` blackhole for `raw.githubusercontent.com`/`github.com`/pypi.org. This is **bypassable** — an agent can dial IPs directly. Confirmed reward-hack cases:
- `diffusers-12982` × opus: fetched fix content from `raw.githubusercontent.com/huggingface/diffusers/...` via inline `urllib.request` (IP-direct, /etc/hosts didn't matter)
- `transformers-35348` × codex: same pattern

v2 (this PR): iptables `OUTPUT DROP` policy + explicit IP allow-list for the small set of hosts an agent legitimately needs:
- Model APIs: `api.anthropic.com`, `api.openai.com`, `router.huggingface.co`
- Agent installers: `deb.debian.org`, `pypi.org`, `files.pythonhosted.org`, `registry.npmjs.org`, GitHub Releases (agent binaries)
- Fix-content hosts (raw.githubusercontent.com, github.com repo pages, direct-IP wheels of the target repo) stay refused at the packet layer

Requires `CAP_NET_ADMIN` in compose. Falls back cleanly to /etc/hosts guard with a stderr warning if iptables is missing.

## Status

- `pipeline_version = "0.9.0"`, class marked `experimental = True`
- 6 of 12 planned M3 gates in place (see header docstring of `pr_to_env.py` for the full checklist); gates 4/6/7/9 are lower-frequency failure modes still on the roadmap
- Reference dataset built: [`AdithyaSK/HF_ML_Bench_v0`](https://huggingface.co/datasets/AdithyaSK/HF_ML_Bench_v0) — 16 oracle-verified tasks across 5 HF ML repos, with an eval matrix across 4 agent tracks (sonnet, opus, codex, qwen)

## Related

PR: (will link once opened)

adithya-s-k:
## Summary

- **New pipeline `pr_to_env`** — the import-shape sibling of `pr_runtime`. Takes an explicit list of PR URLs (`url=…` or `urls_file=…`) and emits one Harbor task per URL, oracle-gated in-loop. Fills the dataset-curation gap where mine-then-filter is the wrong tool. Registered as `experimental=True`.
- **Egress firewall v2** in `_env_guard.py` — replaces the bypassable `/etc/hosts` blackhole with iptables `OUTPUT DROP` + IP allow-list. Closes a confirmed reward-hack channel (opus + codex fetched fix content from `raw.githubusercontent.com` via IP-direct, sidestepping the v1 host guard).
- **In-pipeline gates**: oracle-gate that drops sub-1.0 envs, `keepers.jsonl` ledger, leak-strip v2 (short-SHA + pytest node-id), pyproject.toml sanitize (Dockerfile RUN step for `[tool.pytest]` / `[tool.pytest.ini_options]` collisions on peft).
- **34 tests** added across `test_pipeline_pr_to_env.py` + `test_env_guard.py`; full suite 740/740 green, ruff clean, contract test conforms.

Reference dataset built with this pipeline: [`AdithyaSK/HF_ML_Bench_v0`](https://huggingface.co/datasets/AdithyaSK/HF_ML_Bench_v0) (16 oracle-verified tasks across 5 HF ML repos, with 4-agent eval matrix baked into the README).

Closes #84

## Test plan

- [x] `uv run pytest -q` → 740 passed, 5 skipped
- [x] `uv run ruff check .` → clean
- [x] Contract test: new pipeline conforms to `Pipeline` Protocol
- [x] Reference dataset built end-to-end: 16 tasks emit + oracle-verify + push to HF Hub

## Out of scope

4 lower-frequency M3 gates still pending (parametrization suffix expansion, cross-repo dep pinning, salvage manifest tracking, LF-normalized content hash). Land as follow-ups — pipeline is honest about being `experimental = True`.

Version bump: **not in this PR**. Only after merge + review.


KNambiarDJsc (commented):
Reviewed and tested `da423fa`. This is a substantial, well-documented PR (RFC 0007 already exists on `main` for it) — went through the pipeline, the v2 firewall, and the CI blocker.

**The one thing actually blocking CI is trivial.** All three `Lint (ruff)` failures collapse to one: `ruff format` wants a single trailing blank line removed from the end of `src/repo2rlenv/pipelines/_env_guard.py` (`ruff check` itself is already clean). `uv run ruff format src/repo2rlenv/pipelines/_env_guard.py` fixes it in one commit — `Tests (py3.12/13/14)` are already green and have been since this opened.

**Test suite:** ran the full suite myself rather than trusting the PR description. Linux (WSL, matching CI's `uv sync --group dev --all-extras --frozen`): **739 passed, 6 skipped, 0 failed** — matches your claimed 740/5 (the one extra skip in my run is just `gh not authenticated` in my sandbox). On Windows I saw 8 unrelated failures, all pre-existing Windows-portability gaps already tracked in #130 (bundle-mode checks, symlink privilege, etc.) — nothing here is this PR's fault, noting it only so it isn't mistaken for a regression.

**Two things worth raising on the v2 firewall before merge**, since `_env_guard.py` is this project's actual security boundary against reward-hacking (per `CLAUDE.md`'s "the environment enforces, the prompt never asks" invariant), not just a nice-to-have:

1. **IPv6 is completely unfiltered.** `egress_firewall_entrypoint()` only ever calls `iptables` — `ip6tables` is never touched anywhere in the fragment, and `getent ahostsv4` (grade.py:205 in the diff) explicitly discards any AAAA records it resolves. The default `OUTPUT` policy on the `ip6tables` table is left at whatever the base image ships (normally ACCEPT), so on any container with an IPv6 route, an agent can reach `raw.githubusercontent.com` (or anything else) over its AAAA record and the entire v2 guard — including the exact IP-hardcoded-connection bypass this PR was written to close — is void. Docker's *default* bridge network has IPv6 disabled, which is probably why this hasn't shown up yet, but the PR targets 5 different sandbox providers (Modal/Daytona/E2B/Runloop, per the project's own docs) and some of those enable dual-stack by default. Given the docstring's claim ("Blocks all outbound traffic at the packet layer except a small allow-list") is currently only true for IPv4, I'd treat this as worth closing before relying on it, even if just `ip6tables -P OUTPUT DROP` unconditionally (no allow-list needed there, since none of `DEFAULT_ALLOW_HOSTS` need to be reached over v6 if their v4 addresses are already allowed).
2. **Port 53 is allowed to any destination, not just the container's resolver.** `iptables -A OUTPUT -p udp/tcp --dport 53 -j ACCEPT` has no `-d` restriction, so DNS-tunneling to an attacker-controlled external nameserver (a well-known exfil/infil channel, and squarely in the threat model this PR names — "DoH resolvers bypass it entirely") isn't actually closed, just narrowed to port 53 specifically. Restricting to the container's actual resolver (`-d 127.0.0.11` for Docker's embedded DNS, or read from `/etc/resolv.conf` at entrypoint time) would close this the same way the rest of the allow-list does.

Neither of these undoes the real improvement here over v1 (IPv4 IP-hardcoded bypass is genuinely closed), and I don't have visibility into whether IPv6 is even reachable on the sandbox providers this runs against in practice — flagging them because this is the file the project explicitly trusts to make reward-hacking structurally impossible rather than just discouraged, so it seemed worth having on the record before merge rather than after.

`pr_to_env.py`'s subprocess call (`_run_oracle_gate`) passes an argv list with no `shell=True`, so no injection concern there — spot-checked it since it's the one place PR-URL-derived input reaches a subprocess.

LGTM on the mechanical/functional side once the one-line format fix lands; the two firewall points above are for you and the maintainers to weigh, not blockers I'm asserting.

adithya-s-k:
Closing this since PR-to-env is now covered by Tasksmith and the SWE-gen pipeline on main. Keeping the branch for reference; any remaining firewall improvements can come as a separate focused PR.

PR #91, opened 2026-09-05
Fix pytest parsing of parametrized node IDs containing spaces

### LINKED ISSUE #90: Preserve spaces in parametrized pytest node IDs during parsing and grading
### Description

Both pytest parsers drop verbose result lines or truncate summary node IDs when a parametrized test name contains spaces. Pytest emits these spaces as part of the node ID, so parsing must preserve them to identify tests correctly.

The generation-time parser and the standalone runtime verifier contain the same defect:

- `src/repo2rlenv/log_parsers/pytest_parser.py:35-92`
- `src/repo2rlenv/pipelines/_pr_runtime_verifier.py:74-114`

Their verbose regex captures the name with `\S+`, which cannot span whitespace. Their summary parsing uses `split()` and selects one token as the test name, discarding the rest of a parameter ID containing spaces.

### Reproduction

Run from an environment with Repo2RLEnv installed:

```python
from repo2rlenv.log_parsers.pytest_parser import parse_pytest
from repo2rlenv.pipelines import _pr_runtime_verifier as verifier

verbose = "tests/test_calc.py::test_eval[1 + 1] PASSED [ 50%]\n"
summary = "FAILED tests/test_calc.py::test_eval[1 + 1] - AssertionError: fail\n"

for parser in (parse_pytest, verifier.parse_pytest):
    print(parser(verbose))
    print(parser(summary))
```

Actual output from each parser:

```python
{}
{'tests/test_calc.py::test_eval[1': 'FAILED'}
```

Expected output:

```python
{'tests/test_calc.py::test_eval[1 + 1]': 'PASSED'}
{'tests/test_calc.py::test_eval[1 + 1]': 'FAILED'}
```

This also occurs with real pytest output. A `pytest -v --color=no` run containing two passing cases parametrized with `"1 + 1"` and `"2 * 3"`, plus an ordinary passing test, produced three passing result lines. Both parsers retained only the ordinary test.

### Impact

Generation computes F2P/P2P sets from the parsed pre-fix and post-fix logs in `pipelines/pr_runtime_validate.py`. Missing or truncated identities can omit valid failing-to-passing tests, reject otherwise useful candidates, or merge distinct parameter cases under the same truncated key.

At runtime, grading performs exact node-ID lookups. If a declared F2P test has a full node ID containing spaces while another test makes the parsed status map nonempty, the affected passing test is counted as not passed. With that test as the sole F2P entry, a real all-passing pytest log reproduced `reward = 0.0` and `resolved = False`.

The zero-reward outcome is conditional: when the entire status map is empty and the test command exits successfully, `main()` falls back to `reward = 1.0`, but keeps `resolved = False` when F2P is declared. Also, generated F2P lists are not guaranteed to contain full spaced IDs, because generation itself uses the broken parser.

### Expected behavior

Preserve the complete node ID in both parsers, for verbose and summary output. Extract status/progress and summary diagnostics without treating whitespace inside parameter IDs as a name boundary. Keep the canonical parser and standalone verifier consistent.

Regression coverage should include ordinary names, spaced parameter IDs, distinct IDs sharing a prefix, parameter text containing ` - ` or status words and grading against full node IDs.

abhinavgautam01:
## Summary

Pytest node IDs containing spaces were dropped from verbose output or truncated in short summaries. This could omit tests from F2P/P2P discovery and incorrectly penalize passing tests during grading.

Update both the generation-time parser and standalone runtime verifier to preserve complete parameter IDs while separating pytest statuses, progress indicators and diagnostic messages. Preserve folded skip handling and last-write-wins behavior.

Add 72 regression cases covering spaces, repeated whitespace, embedded dashes and status words, brackets, skip reasons and distinct parameter identities. Integration tests exercise actual pytest output, F2P/P2P discovery and standalone verifier grading.

Fixes #90.

## Validation

- Python 3.12, 3.13, and 3.14: **788 passed, 5 skipped** on each.
- `ruff check .` and `ruff format --check .` passed.
- Source distribution and wheel builds passed.
- Clean wheel installation and CLI version check passed.

### DIFF
--- a/src/repo2rlenv/log_parsers/pytest_parser.py
+++ b/src/repo2rlenv/log_parsers/pytest_parser.py
@@ -30,9 +30,15 @@ TestStatus = Literal["PASSED", "FAILED", "SKIPPED", "ERROR"]
 _STATUSES = ("PASSED", "FAILED", "SKIPPED", "ERROR")
 
 # Verbose progress format: `tests/foo.py::test_x PASSED  [12%]` or `... FAILED`
-# Test names contain alphanumerics, _, /, ., ::, [, ], -, +
-# We require AT LEAST ONE non-whitespace token followed by whitespace + STATUS.
-_VERBOSE_RE = re.compile(r"^(?P<name>\S+)\s+(?P<status>PASSED|FAILED|SKIPPED|ERROR)\b")
+# Anchor the status to pytest's trailing reason/progress fields, so spaces and
+# status words inside a parameter ID remain part of the name.
+_VERBOSE_RE = re.compile(
+    r"^(?P<name>.+?(?:\[.*?\])?)\s+(?P<status>PASSED|FAILED|SKIPPED|ERROR)"
+    r"(?:\s+\(.*\))?(?:\s+\[\s*\d+%\])?$"
+)
+# Consume a bracketed parameter suffix before looking for the diagnostic dash.
+# Lazy matching keeps brackets in the diagnostic from becoming part of the ID.
+_SUMMARY_NAME_RE = re.compile(r"^(?P<name>.+?(?:\[.*?\])?)(?: - .*)?$")
 
 
 def parse_pytest(log: str) -> dict[str, TestStatus]:
@@ -41,8 +47,8 @@ def parse_pytest(log: str) -> dict[str, TestStatus]:
     Notes:
       - Last-write-wins. A test that appears in both progress AND summary
         ends up as whichever appeared last (typically summary, which is fine).
-      - SKIPPED lines like `SKIPPED [1] tests/foo.py:42` have the [N] count
-        prefix stripped so the test name is the third token, not '[1]'.
+      - SKIPPED lines like `SKIPPED [1] tests/foo.py:42` record the file
+        location rather than the count prefix or skip reason.
       - Lines like `FAILED tests/foo.py::test_x - AssertionError: ...`
         get the dash chunk stripped to keep the test name clean.
       - Returns an empty dict for empty/malformed input. Caller decides what
@@ -57,28 +63,22 @@ def parse_pytest(log: str) -> dict[str, TestStatus]:
             continue
 
         # --- format (2): summary lines (STATUS first) ---
-        # Has to be checked BEFORE the verbose regex because a summary line
-        # like "PASSED tests/foo.py::test_a" would also match the verbose
-        # pattern (with name="PASSED" — wrong).
+        # Handle the leading status separately from the full node ID.
         leading_status: TestStatus | None = None
         for st in _STATUSES:
             if line.startswith(st + " ") or line == st:
                 leading_status = st  # type: ignore[assignment]
                 break
         if leading_status is not None:
-            work = line
-            if leading_status == "FAILED" and " - " in work:
-                work = work.split(" - ", 1)[0]
-            tokens = work.split()
-            if len(tokens) < 2:
-                continue
-            test_name = tokens[1]
-            # SKIPPED [N] file:line  → the [N] is a count, real name is tokens[2]
-            if test_name.startswith("[") and test_name.endswith("]"):
-                if len(tokens) < 3:
+            work = line[len(leading_status) :].strip()
+            if leading_status == "SKIPPED" and re.match(r"^\[\d+\](?:\s|$)", work):
+                # Folded skips report a file location, not a parametrized ID.
+                tokens = work.split(maxsplit=2)
+                if len(tokens) < 2:
                     continue
-                test_name = tokens[2]
-            out[test_name] = leading_status
+                out[tokens[1]] = leading_status
+            elif m := _SUMMARY_NAME_RE.match(work):
+                out[m.group("name")] = leading_status
             continue
 
         # --- format (1): verbose progress (NAME first, STATUS after) ---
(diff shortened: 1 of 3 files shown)


KNambiarDJsc (commented):
Thanks for this. It's the more complete of the two fixes for #90. I checked it against real pytest 9.0.3 output rather than hand-written lines. I recorded each test's outcome through a `pytest_runtest_logreport` hook as ground truth, then parsed the terminal log. The IDs covered spaces, ` - `, status words, and unmatched `[` / `]`.

With `-v`, `-v -rA`, `-vv -rA` and `console_output_style=classic`, both parsers get 17/17. The canonical parser and the standalone verifier always agree. The test that runs the shipped `verifier.py` in a subprocess is a great addition.

Two things I'd fix before merge:

**1. Regression: `console_output_style = count` / `times`.** The verbose regex is `$`-anchored and only allows a `[ NN%]` suffix. With the `count` and `times` styles, pytest ends the line differently, so every verbose line stops matching. `main` still parses these for ordinary names:

```
test_cases.py::test_pass[plain] PASSED        [ 8/17]
test_cases.py::test_pass[plain] PASSED        415.6us
```

In those styles only the `FAILED` short-summary lines survive, and every `PASSED` / `SKIPPED` test disappears (7/17 correct). `validate_pr` only records F2P when the post-fix status is `PASSED`, so any repo with one of these styles in its pytest config would produce no candidates.

**2. Quadratic backtracking on long non-test lines.** `.+?(?:\[.*?\])?` is tried on every log line, including assertion diffs. On a single line of bracket-heavy text (`E assert [0, [x]], [1, [x]], ...`), `parse_pytest` took:

| line length | time |
|---|---|
| 5k chars | 0.036s |
| 24k chars | 0.61s |
| 50k chars | 2.5s |

That's roughly 4× per doubling. `-vv` diffs can be that long, generation parses logs up to 5 MB, and the verifier runs under the task's `timeout_sec`.

A small change fixes both. Accept pytest's other progress suffixes (`_pytest/terminal.py`: `[ 1/17]`, `[ 17 / 17 ]`, `format_node_duration`). Keep `main`'s `::` / `.py` heuristic, but check it on the first token *before* running the regex. Node IDs keep `::` in the first token even when the parameter contains spaces. Same edit in `_pr_runtime_verifier.py`:

```python
_PROGRESS = (
    r"\[\s*\d+%\]|\[\s*\d+\s*/\s*\d+\s*\]"  # percent / count styles
    r"|\d+(?:\.\d+)?(?:us|ms|s)|\d+m \d+s|\d+h \d+m"  # times style
)
_VERBOSE_RE = re.compile(
    r"^(?P<name>.+?(?:\[.*?\])?)\s+(?P<status>PASSED|FAILED|SKIPPED|ERROR)"
    rf"(?:\s+\(.*\))?(?:\s+(?:{_PROGRESS}))?$"
)
...
        # --- format (1): verbose progress (NAME first, STATUS after) ---
        head = line.split(None, 1)[0]
        if "::" not in head and not head.endswith(".py"):
            continue
        m = _VERBOSE_RE.match(line)
```

With that change:
- Real-pytest runs: 17/17 in all six output styles.
- 50k-char line: 0.000s.
- 100k-line realistic log: 0.25s (0.41s before).
- Your 72 tests plus `test_log_parsers*`, `test_pr_runtime_verifier` and `test_grading`: all 127 pass.

It would be worth adding `count` / `times` cases to `test_real_pytest_logs_preserve_oracle_and_standalone_grading`.

Non-blocking: a parameter containing `] - `, e.g. `test_x[a] - b]`, is inherently ambiguous in ` - ` summary lines. It yields an extra `test_x[a]` key, but the verbose line still records the right ID, so I don't think it needs handling.

abhinavgautam01:
Thanks @KNambiarDJsc for the detailed review, confirmed and fixed both findings in both parsers. Added count/times suffix coverage, real-pytest checks against report-hook outcomes across output styles and verbosity levels and a large assertion-diff regression test. Tests pass on Python 3.12–3.14; lint and formatting are clean. Left the non-blocking `] - ` summary ambiguity unchanged as suggested.

KNambiarDJsc:
Thanks for the quick turnaround. I re-ran my checks on ff9cedc, and everything from the review is resolved:

- **Real pytest 9.0.3.** I compared against `pytest_runtest_logreport` outcomes under `-v`, `-v -rA`, `-vv -rA`, and the `classic`, `count` and `times` styles. All 19/19 are correct in each, and the canonical parser matches the standalone verifier every time. The IDs covered spaces, ` - `, status words, unmatched `[`, and `::` inside the failure message.
- **Performance.** A 50k-char bracket-heavy line now takes 0.000s (it was 2.5s). A 448k-char line full of status words takes 0.000s, and a 100k-line log takes 0.25s.
- **Durations.** `2m 5s` and `1h 3m` durations (tests ≥ 60s in `times` style) parse correctly.
- **CI-equivalent run.** On Linux with py3.12, `ruff check`, `ruff format --check` and the full suite pass (816 passed). The #92 test file also passes against this branch (88/88).

LGTM from my side. Happy for a maintainer to take it from here.

adithya-s-k (approved):

adithya-s-k:
thanks @abhinavgautam01, merged! appreciate the parser fixes and the real pytest checks. thanks @KNambiarDJsc for reviewing too!

Issue #65, opened 2026-06-15
GitLab: support private-repo MR diff-fetch


adithya-s-k:
Follow-up from #64 (GitLab MR mining).

`pr_diff` / `pr_runtime` work on **public** gitlab.com repos. **Private** repos are not yet supported: `gitlab.fetch_pr_diff` fetches the MR diff from the non-API web route `gitlab.com/<owner>/<name>/-/merge_requests/<iid>.diff`, which the `PRIVATE-TOKEN` header does **not** authenticate (it's not under `/api/v4`). So on a private repo the list/issue API calls authenticate fine, but the diff fetch would 401.

### Options
1. Reconstruct the unified diff from the API: `GET /api/v4/projects/:id/merge_requests/:iid/diffs` (or `/changes`, already called in `list_merged_prs`) returns per-file `diff` bodies — prepend `diff --git` / `---` / `+++` headers to make it `git apply`-able. Header-authed, works for private.
2. Use `?private_token=` / `?access_token=` on the web `.diff` route — simpler but puts the token in the URL (must scrub from error logs).

Option 1 is cleaner and consistent with the rest of the client. Should ship with a live private-repo test.

adithya-s-k:
Closing as not currently planned. GitLab MR mining shipped in #62/#64, but private-repo MR diff-fetch specifically wasn't taken up this cycle. Reopen if it becomes a priority.

PR #94, opened 2026-09-06
Fix GitLab clone URLs and authentication in pr_diff Dockerfiles

### LINKED ISSUE #93: Use the source repository host when emitting pr_diff Dockerfiles for GitLab
### Description

`pr_diff` supports GitLab merge-request mining through the source-provider abstraction, but its emitted Dockerfile reconstructs the clone URL as a GitHub URL:

```python
repo_url = f"https://github.com/{owner}/{name}.git"
```

This happens in `src/repo2rlenv/pipelines/pr_diff.py:574`, inside `PRDiffPipeline._build_task()`, despite `self.input.repo.url` retaining the GitLab source URL.

The Dockerfile builder also hardcodes `ARG GITHUB_TOKEN` and performs token injection only for `https://github.com/` (`pr_diff.py:208-239`). Supplying a correct GitLab URL directly to the builder still does not enable GitLab token authentication.

### Reproduction

Construct a task through `PRDiffPipeline._build_task()` using:

```python
RepoSpec(
    url="https://gitlab.com/python-devs/importlib_resources",
    access="public",
)
```

Use any suitable `PullRequestSummary` fixture and nonempty diff with `PRDiffOptions(emit_harbor_env=True)`. Inspect `task.environment_dockerfile`.

The generated Dockerfile contains:

```dockerfile
ARG GITHUB_TOKEN=
RUN if [ -n "$GITHUB_TOKEN" ]; then \
        git clone --filter=blob:none https://x-access-token:${GITHUB_TOKEN}@github.com/python-devs/importlib_resources.git /workspace; \
    else \
        git clone --filter=blob:none https://github.com/python-devs/importlib_resources.git /workspace; \
    fi \
 && git -C /workspace remote set-url origin https://github.com/python-devs/importlib_resources.git
```

This output was reproduced locally through the task builder. A Docker build was not needed to establish the incorrect destination and no specific live clone error is asserted here.

### Impact

The environment clones a different host from the repository that supplied the merge request and base commit. Builds can fail because the GitHub repository or commit is unavailable. If a same-named repository exists on GitHub, the build still targets the wrong source.

This affects public GitLab repositories with environment emission enabled, the default setting. It does not require private-repository support to reproduce.

### Expected behavior

Build the clone and origin URLs from the actual source repository URL, preserving its host and path. Keep GitHub behavior working and make build-time authentication source-aware, including GitLab token handling where supported. Preserve the existing credential cleanup behavior.

Add task-builder regression coverage that checks a GitLab input emits GitLab clone and origin URLs, plus focused Dockerfile-builder coverage for provider-specific authentication.

abhinavgautam01:
Fixes #93

## Summary

GitLab merge requests were mined through the correct provider, but their emitted `pr_diff` Dockerfiles cloned a hardcoded GitHub URL. This could fail the environment build or target a different repository.

Build clone URLs from the source repository URL, preserving the host and full path. Normalize supported SSH-style inputs to HTTPS and select `GITHUB_TOKEN` or `GITLAB_TOKEN` using the existing authentication helper.

Quote token expansion, remove credentials supplied in source URLs and retain the clean origin reset after cloning. Update documentation to describe provider-specific authentication.

## Validation

Add 15 regression cases covering GitHub compatibility, GitLab URLs, nested paths, SSH-style inputs, credential stripping and public/authenticated clone commands.

The shell tests execute the generated clone commands with a fake Git executable and dummy tokens, checking argument handling and origin cleanup without network access.

- Python 3.12, 3.13 and 3.14: **731 passed, 5 skipped** on each.
- Ruff lint, formatting and whitespace checks passed.
- Source distribution and wheel builds passed.
- Clean wheel installation with locked dependencies, CLI version check and dependency compatibility check passed.

## Scope

Private GitLab MR diff fetching remains unsupported under #65. This change fixes emitted clone URLs and build-time authentication; it does not establish end-to-end private GitLab mining support.


KNambiarDJsc (commented):
I reviewed and tested 719238d:

- **CI-equivalent run on Linux:** ruff is clean, and the full suite passes on py3.12 and py3.14 (730 passed).
- **Generated Dockerfiles, `main` vs this PR:**
  - `--repo pallets/click`: the only change is the quoted token expansion, `x-access-token:"${GITHUB_TOKEN}"@github.com/...`. The shell treats it the same, so GitHub tasks are unaffected.
  - `--repo https://gitlab.com/python-devs/importlib_resources`: `main` emitted `ARG GITHUB_TOKEN` and cloned `https://github.com/python-devs/importlib_resources.git`. With this PR it emits `ARG GITLAB_TOKEN`, clones through `oauth2:"${GITLAB_TOKEN}"@gitlab.com/...`, and uses the correct clean URL.
- **The quoting tests are thorough.** They run the real `RUN` body through `/bin/sh` with a fake `git`, including a token containing spaces, `*`, `$HOME` and `$(false)`.

LGTM. Two non-blocking follow-ups:

1. **`docs/reference/AUTH.md` is still GitHub-only.** Its "Private repos at task **build** time" section says the Dockerfile declares `ARG GITHUB_TOKEN=` and clones through `x-access-token:<token>@github.com/...`, so a consumer following it for a GitLab task would pass the wrong build arg. Mentioning `GITLAB_TOKEN` and `oauth2:` there would match the updated `pr_diff.md`.
2. **Windows test runs.** The 6 `test_clone_shell_uses_correct_token_and_scrubs_origin` cases fail on Windows with `FileNotFoundError`, because there's no `/bin/sh`. CI is Linux-only, so there's no CI impact. Adding `pytest.mark.skipif(not Path("/bin/sh").exists(), reason="needs a POSIX shell")` would keep the suite usable there; #108 tracks the other Windows issues.

adithya-s-k (approved):

adithya-s-k:
thanks @abhinavgautam01, merged! appreciate the GitLab clone and auth fixes, especially the coverage around credentials.

PR #97, opened 2026-09-07
Support nested task dataset layouts when pushing to Hub

### LINKED ISSUE #96: Accept nested tasks/ dataset layouts when pushing to Hugging Face Hub
### Description

Dataset push currently discovers only immediate child directories containing `task.toml`. It fails when given a dataset root whose tasks live under `tasks/`, even though Repo2RLEnv publishes that layout and its pull implementation recognizes it.

Affected discovery paths:

- `src/repo2rlenv/hub.py:473-498`, inside `push_to_hub()`.
- `src/repo2rlenv/registry/integration.py:70-79`, `_list_task_dirs()`, used by image preparation.

Both should recognize:

```text
dataset/
  task-a/task.toml
```

and:

```text
dataset/
  README.md
  registry.json
  tasks/
    task-a/task.toml
```

### Reproduction

Given the second layout:

```python
from pathlib import Path
from repo2rlenv.registry.integration import _list_task_dirs

print(_list_task_dirs(Path("dataset")))
# Actual: []

print(_list_task_dirs(Path("dataset/tasks")))
# Actual: [PosixPath('dataset/tasks/task-a')]
```

Calling `push_to_hub()` on `dataset/` raises:

```text
RuntimeError: no Harbor tasks found in .../dataset
```

The push failure was reproduced with HF authentication/API calls and image preparation mocked, so the reproduction performed no external upload.

### Impact

Users cannot directly push the root of a git-cloned dataset or raw Hub snapshot containing `tasks/<id>/task.toml`. They must rearrange the files or point push at the `tasks/` subdirectory.

This is a direct-input compatibility gap. The normal `repo2rlenv pull` → `repo2rlenv push` workflow does not inherently fail: pull already flattens nested task directories into its output directory.

### Expected behavior

Use consistent task discovery for both image preparation and Hub staging, accepting flat and nested layouts. Continue excluding hidden directories and staging artifacts. Define deterministic behavior when both layouts appear and handle duplicate task IDs explicitly rather than silently overwriting tasks.

Add coverage for both supported layouts, an empty dataset, hidden directories and discovery consistency between preparation and upload staging.

karatarassul4-max:
Fixes #96

## Summary

* reuse `_list_task_dirs()` in `push_to_hub()` so Hub staging and image preparation share the same discovery logic
* support both flat task layouts and the canonical `tasks/<id>/task.toml` layout
* keep hidden directories excluded and preserve deterministic ordering
* reject duplicate task IDs across flat and nested layouts instead of silently choosing one copy
* add regression coverage for nested discovery, mixed layouts, duplicate IDs, and nested `push_to_hub()` metadata discovery

## Validation

* `uv run pytest -q tests/registry/test_integration.py tests/test_hub_metadata_read.py` → 34 passed
* `uv run ruff check src/repo2rlenv/registry/integration.py src/repo2rlenv/hub.py tests/registry/test_integration.py tests/test_hub_metadata_read.py` → passed
* `uv run ruff format --check src/repo2rlenv/registry/integration.py src/repo2rlenv/hub.py tests/registry/test_integration.py tests/test_hub_metadata_read.py` → 4 files already formatted
* `git diff --check` → clean
* full `uv run pytest -q` on Windows: 712 passed, 6 skipped, 7 platform-specific failures related to executable bits, path separators, and Windows local-path parsing; none are in files changed by this PR

### DIFF
--- a/src/repo2rlenv/hub.py
+++ b/src/repo2rlenv/hub.py
@@ -428,7 +428,7 @@ def push_to_hub(
     """
     from huggingface_hub import HfApi
 
-    from repo2rlenv.registry.integration import prepare_dataset_for_push
+    from repo2rlenv.registry.integration import _list_task_dirs, prepare_dataset_for_push
 
     token = resolve_hf_token(auth)
     if not token:
@@ -470,12 +470,8 @@ def push_to_hub(
     first_metadata: dict[str, Any] = {}
     source_repos: set[str] = set()
     has_environment = False
-    for child in sorted(local_dataset_dir.iterdir()):
-        if not child.is_dir() or child.name.startswith("."):
-            continue
+    for child in _list_task_dirs(local_dataset_dir):
         toml_path = child / "task.toml"
-        if not toml_path.exists():
-            continue
         meta = _read_task_metadata(toml_path)
         if not first_metadata:
             first_metadata = meta
--- a/src/repo2rlenv/registry/integration.py
+++ b/src/repo2rlenv/registry/integration.py
@@ -68,14 +68,30 @@ _FROM_LINE_RE = re.compile(r"^(\s*FROM\s+)(\S+)", re.IGNORECASE | re.MULTILINE)
 
 
 def _list_task_dirs(local_dir: Path) -> list[Path]:
-    """Return task directories (each containing task.toml), sorted, no hidden."""
-    out = []
-    for child in sorted(local_dir.iterdir()):
-        if not child.is_dir() or child.name.startswith("."):
-            continue
-        if (child / "task.toml").exists():
-            out.append(child)
-    return out
+    """Return task directories from flat or ``tasks/`` dataset layouts.
+
+    A dataset may contain tasks directly under its root or under the canonical
+    Hub ``tasks/`` directory. Both layouts can coexist, but duplicate task IDs
+    are rejected rather than silently choosing one copy.
+    """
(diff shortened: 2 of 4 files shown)


KNambiarDJsc (commented):
I reviewed and tested 92b5138:

- **CI-equivalent run on Linux:** ruff is clean, and the full suite passes on py3.12 and py3.14 (719 passed).
- **End-to-end on a real nested dataset.** I copied three tasks from the published `AdithyaSK/repo2rlenv-pr-diff` dataset in its `tasks/<id>/` layout. I also added a root `README.md` and `registry.json`, as `huggingface-cli download` leaves them. Then I ran `push_to_hub` with `HfApi` mocked.
  - On `main` it raises `RuntimeError: no Harbor tasks found`.
  - With this PR, all 3 tasks are staged under `tasks/`, and the generated `manifest.json` has the same 3 rows.
  - Nothing ends up double-nested as `tasks/tasks/`.
  - The push runs through image preparation without errors; these tasks ship `environment/Dockerfile`s.

LGTM.

One non-blocking nit: `hub.py` now imports the private `_list_task_dirs` from `registry.integration`. If both modules are meant to share discovery, dropping the underscore would make that dependency explicit.

adithya-s-k (approved):

adithya-s-k:
thanks @karatarassul4-max, merged! appreciate the nested task layout support and the duplicate ID checks.

PR #99, opened 2026-09-08
fix(commit_runtime): use the real host in the dataset reference URL

### LINKED ISSUE #98: commit_runtime's dataset reference URL hardcodes github.com for GitLab-sourced repos
`commit_runtime.py`'s `_build_task()` builds the dataset's `reference` provenance link as
`f"https://github.com/{owner}/{name}/commit/{commit.sha}"` unconditionally. `RepoSpec.owner_name`
strips both the `github.com` and `gitlab.com` prefixes down to bare `(owner, name)`, so a
GitLab-sourced repo loses its host there, and this line always writes it back as `github.com`.
Any `commit_runtime` dataset built from a GitLab repo ends up with a reference link pointing at
a `github.com` URL that doesn't exist.

`sources.py` documents `commit_runtime` as one of the git/source-based pipelines that work on
any source (GitHub, GitLab, local), so this isn't a scope gap, just a missed case: the
`reference` field never looks at `source_kind`. Local checkouts have the same problem for a
different reason: there's no real host to link to, and the code links to `github.com` anyway.

Repro: build a `RepoSpec` with a `gitlab.com` URL and call `_build_task()`; the emitted
`repo2env["reference"]` is a github.com link that 404s. Same story for a local checkout.

Have a fix (route the host through `self.input.repo.source_kind`, drop `reference` for local
since there's no public URL) plus tests, opening a PR.

AmirF194:
## Summary

- `_build_task()` always wrote the dataset's `reference` link as
  `https://github.com/{owner}/{name}/commit/{sha}`, even when the source repo is GitLab or a
  local checkout. `owner_name` strips the host, so nothing downstream knew which one it was.
- Route the host through `self.input.repo.source_kind` instead: `github.com` for GitHub,
  `gitlab.com` for GitLab, and drop `reference` entirely for a local checkout, since there's no
  real URL to point at there and TOML has no null for `tomli_w` to write anyway.

## Test plan

- [x] Added `test_build_task_reference_host_matches_source`, parametrized over GitHub/GitLab/local.
      The GitHub case passes on main (behavior unchanged); GitLab and local both fail on main
      and pass on the branch.
- [x] Added `test_build_task_local_source_writes_a_loadable_task_toml`, driving the real
      `write_harbor_task` emitter for the local case and loading the result back with `tomllib`.
      Caught a real bug in my first pass: setting `reference` to `None` instead of omitting the
      key crashes `tomli_w.dumps` (TOML has no null).
- [x] Full suite: 719 passed, 6 skipped, on 3.12 and 3.14.
- [x] `ruff check` / `ruff format --check`: clean.

## Out of scope

`code_instruct.py` and `equivalence_tests.py` build a similar `blob/...#Lx-Ly` reference the
same way and have the identical hardcoded-host bug. Same root cause, different files; happy to
send a follow-up if that's useful, didn't want to bundle three pipelines into one PR.

Closes #98

### DIFF
--- a/src/repo2rlenv/pipelines/commit_runtime.py
+++ b/src/repo2rlenv/pipelines/commit_runtime.py
@@ -65,7 +65,7 @@ from repo2rlenv.pipelines.pr_runtime import (
     targeted_test_cmds_for_pr,
 )
 from repo2rlenv.provider import provider_for
-from repo2rlenv.sources import Capability, capabilities_for
+from repo2rlenv.sources import Capability, SourceKind, capabilities_for
 from repo2rlenv.spec.input import GenerationInput, PipelineName
 from repo2rlenv.spec.options import CommitRuntimeOptions
 
@@ -545,6 +545,13 @@ class CommitRuntimePipeline:
         # commit_runtime task ID convention: <owner>__<repo>-<sha12>
         # (analogous to pr_runtime's <owner>__<repo>-<pr_number>)
         task_id = f"{owner}__{name}-{commit.sha[:12]}"
+        source_kind = self.input.repo.source_kind
+        commit_reference = (
+            f"https://{'gitlab.com' if source_kind is SourceKind.GITLAB else 'github.com'}"
+            f"/{owner}/{name}/commit/{commit.sha}"
+            if source_kind is not SourceKind.LOCAL
+            else None
+        )
 
         eval_script = build_eval_script(
             base_commit=commit.parent_sha,
@@ -577,7 +584,9 @@ class CommitRuntimePipeline:
             "pipeline_version": "0.8.3",
             "repo": f"{owner}/{name}",
             "ref": commit.parent_sha,
-            "reference": f"https://github.com/{owner}/{name}/commit/{commit.sha}",
+            # Omitted, not None, for a local checkout: TOML has no null and
+            # tomli_w rejects it outright.
+            **({"reference": commit_reference} if commit_reference is not None else {}),
             "source_access": self.input.repo.access,
             "built_at": datetime.now(UTC).isoformat(),
             **({"synthesis_llm": self.input.llm.qualified_name} if self.input.llm else {}),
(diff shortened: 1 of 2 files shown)


KNambiarDJsc (commented):
I reviewed and tested fa06f9e:

- **CI-equivalent run on Linux:** `ruff check` and `ruff format --check` are clean, and the full suite passes on py3.12 and py3.14 (719 passed).
- **Omitting `reference` for local sources is safe.** Its only reader, `hub.py`'s manifest builder, uses `r2e.get("reference")`.
- **The GitLab link resolves.** `https://gitlab.com/<ns>/<project>/commit/<sha>` returns a 301 redirect to the canonical `/-/commit/<sha>`. I checked this against a real `python-devs/importlib_resources` commit.

LGTM. Two optional, non-blocking suggestions:

1. **Nested GitLab groups.** `RepoSpec.owner_name` keeps only the last two path segments. For `https://gitlab.com/gitlab-org/api/client-go` it returns `("api", "client-go")`, so the reference becomes `https://gitlab.com/api/client-go/commit/...`, which 404s. `commit_runtime` mines with plain `git` and only uses the GitLab API for optional issue lookup, so it otherwise works on nested-group repos. Building the link from the normalized `self.input.repo.url` would keep the full namespace; the `git@gitlab.com:` form would need converting to https first.
2. **Canonical GitLab path.** Using `/-/commit/` for GitLab links avoids the redirect.

#113 reports the same bug in `code_instruct` and `equivalence_tests`. A small shared helper for building the reference URL would let that follow-up reuse this logic.

AmirF194:
Thanks for testing it end to end, including the real GitLab redirect check.

Pushed 9781137 for the canonical /-/commit/ path, tests updated to match.

Leaving nested GitLab groups out of this PR: fixing it properly means building the reference from the normalized repo URL instead of the stripped (owner, name) pair, which touches the same code as code_instruct.py and equivalence_tests.py. #113 already tracks that as one problem across all three files, so a shared helper belongs there rather than as a one-off in this PR.

adithya-s-k (approved):

adithya-s-k:
thanks @AmirF194, merged! appreciate the fix and the GitLab follow-up. keeping #113 open for the remaining reference URL cases.

PR #104, opened 2026-09-12
cli: add --llm-endpoint/--llm-key-env, keyless self-hosted LLMs

### LINKED ISSUE #103: `--llm-endpoint` is documented but does not exist, and self-hosted providers are rejected before LiteLLM is called
### Description

`docs/reference/BOOTSTRAP.md` has advertised `--llm-endpoint` and `--llm-key-env` for self-hosted vLLM / Ollama since v0.8.1, but neither flag exists in `cli.py`. `LLMSpec.endpoint` / `api_key_env` are honoured by `llm._do_complete`, so the only way to reach them today is a `--config` file.

Independently, `auth.resolve_llm_api_key()` only knows five hosted providers. For anything else (`hosted_vllm/…`, `ollama/…`, Bedrock, Vertex, …) it returns `None` and `_do_complete` raises `no API key resolved` before LiteLLM is ever called. LiteLLM itself handles those providers fine (it substitutes a placeholder key for vLLM), so the `Ollama / vLLM` support claimed in `CLAUDE.md` is blocked entirely by our pre-check.

### Reproduction

`vllm serve Qwen/Qwen3.5-4B --port 8000`, no API keys in the environment.

```bash
repo2rlenv generate --repo pallets/click --pipeline commit_runtime \
  --llm hosted_vllm/Qwen/Qwen3.5-4B --llm-endpoint http://127.0.0.1:8000/v1 --out ./out
# repo2rlenv: error: unrecognized arguments: --llm-endpoint http://127.0.0.1:8000/v1
```

With the endpoint moved into a config file instead:

```python
from repo2rlenv.llm import complete
from repo2rlenv.spec.input import LLMSpec

ep = "http://127.0.0.1:8000/v1"
complete(LLMSpec(provider="hosted_vllm", model="Qwen/Qwen3.5-4B", endpoint=ep), user="ping")
# RuntimeError: no API key resolved for provider 'hosted_vllm'. Set the provider-default env var.
complete(LLMSpec(provider="openai", model="Qwen/Qwen3.5-4B", endpoint=ep), user="ping")
# RuntimeError: no API key resolved for provider 'openai'. Set the provider-default env var.
```

Raw LiteLLM against the same server works without a key:

```python
import litellm
litellm.completion(model="hosted_vllm/Qwen/Qwen3.5-4B", api_base=ep, messages=[{"role": "user", "content": "ping"}])
# OK
```

The only working route today is `openai/<model>` + endpoint in a config file + `OPENAI_API_KEY` set to any dummy value, which nothing documents.

### Expected behavior

- `--llm-endpoint URL` and `--llm-key-env VAR` exist on `generate` and `bootstrap`, as documented.
- A self-hosted endpoint needs no API key. Providers LiteLLM resolves itself (`hosted_vllm`, `ollama`, Bedrock, …) are not gated by our resolver.
- The five hosted providers we do resolve still fail fast, naming the env var.

Two related things worth handling in the same change: `.env.example` ships blank `OPENAI_API_KEY=` / `ANTHROPIC_API_KEY=` lines that `load_dotenv` turns into empty strings, and a real `OPENAI_API_KEY` in the environment should not be forwarded as a Bearer token to an arbitrary `--llm-endpoint` host.

### Impact

Anyone with a local GPU cannot use their own model for bootstrap or synthesis, despite three doc pages saying they can. It also blocks the cost-free path for iterating on pipelines.

I have a fix with tests ready and will open a PR.

surajsharan:
## Summary

- Add the `--llm-endpoint` / `--llm-key-env` flags to `generate` and `bootstrap`. Both have been documented in `BOOTSTRAP.md` since v0.8.1 but never existed; the spec fields were only reachable via `--config`.
- Stop rejecting providers our key resolver doesn't know. `hosted_vllm/`, `ollama/`, Bedrock, Vertex, … are now left to LiteLLM's own credential lookup; the five hosted providers in `auth.LLM_KEY_ENV_DEFAULTS` still fail fast with the env var named.
- Make key precedence explicit and safe: `--llm-key-env` always wins and never falls through to another provider's key; with `--llm-endpoint` the provider-default key is never forwarded to the custom server (`openai/` gets a placeholder, `hosted_vllm/` honours `HOSTED_VLLM_API_KEY`); blank env values (as shipped in `.env.example`) count as unset.
- Small related fixes surfaced while testing: `--llm-fallback` now composes with a `--config` file like its siblings; `--llm` over a config whose `llm:` block names a different model drops that block's stale `endpoint` / `api_key_env`; a typo'd provider fails before clone / image pull instead of at the first agent turn; a config with no `llm:` block reports field errors instead of a traceback.

## Test plan

- [x] `uv run pytest -q` — 752 passed, 5 skipped (26 new tests; the new LLM tests stub `litellm` via `sys.modules`, so the suite does not pay its import cost)
- [x] `uv run ruff check .` and `uv run ruff format --check .` clean
- [x] `uv run pytest tests/test_pipeline_contract.py` unchanged and green
- [x] `mkdocs build` clean apart from the pre-existing draft-RFC link warnings
- [x] Live against vLLM 0.20.2 serving `Qwen/Qwen3.5-4B` (RTX 3090), LiteLLM 1.83.14, no keys in the environment:
  - `hosted_vllm/…` + `--llm-endpoint` and `openai/…` + `--llm-endpoint` both complete
  - With a logging proxy in front of vLLM: `openai/` sends `Bearer EMPTY` even when `OPENAI_API_KEY` is set, and sends the real key only with `--llm-key-env OPENAI_API_KEY`; `hosted_vllm/` forwards `HOSTED_VLLM_API_KEY` when set and LiteLLM's own placeholder otherwise
  - `repo2rlenv bootstrap --llm hosted_vllm/… --llm-endpoint …` passes flag parsing and the provider preflight and stops at the Docker boundary on a machine without Docker; `--llm antropic/…` fails in the preflight before any work
  - The `commit_runtime` instruction-synthesis prompt run through the local model on three `pallets/click` commits produced statements that pass the pipeline's own gate

## Out of scope

- A full bootstrap + `harbor run -a oracle` cycle on a self-hosted model (needs Docker; will report separately once run).
- Adding more providers to `LLM_KEY_ENV_DEFAULTS`. Anything not listed is delegated to LiteLLM, which already knows the env var names.
- Cost tracking for self-hosted models. `litellm.completion_cost` has no price entry for them, so they count as $0 and `max_llm_spend_usd` never trips; documented in `BOOTSTRAP.md`.

Closes #103

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -229,7 +229,10 @@ shell-out); gitlab → `repo.auth_token_env` then `$GITLAB_TOKEN`; github →
 
 HF Hub auth uses `huggingface_hub`'s own resolution
 (`~/.cache/huggingface/token` or `$HF_TOKEN`). LLM keys come from provider-default
-env vars (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, …) via `auth.resolve_llm_api_key()`.
+env vars (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, …) via `auth.resolve_llm_api_key()` —
+only for the providers in `auth.LLM_KEY_ENV_DEFAULTS`. Anything else is left to LiteLLM's
+own lookup, and a self-hosted `LLMSpec.endpoint` (`--llm-endpoint`) needs no key at all
+(`llm._resolve_api_key`).
 
 Registry push credentials resolve from explicit env vars first: GHCR reads
 `GHCR_TOKEN` / `GITHUB_TOKEN` (needs a one-time
--- a/docs/reference/API.md
+++ b/docs/reference/API.md
@@ -85,7 +85,7 @@ response = complete(
 print(response.content)
 ```
 
-Single-shot LiteLLM call. Honors `spec.endpoint` for self-hosted backends; auto-points HF provider at `https://router.huggingface.co/v1`.
+Single-shot LiteLLM call. Honors `spec.endpoint` for self-hosted backends — no key required, and the provider-default key is never forwarded there unless `spec.api_key_env` names it; auto-points HF provider at `https://router.huggingface.co/v1`. Providers outside `auth.LLM_KEY_ENV_DEFAULTS` resolve their credentials inside LiteLLM. `check_provider(spec)` fails fast on an unknown provider anywhere in the fallback chain.
 
 ## `repo2rlenv.reward`
 
--- a/docs/reference/AUTH.md
+++ b/docs/reference/AUTH.md
@@ -107,7 +107,9 @@ LiteLLM resolves provider keys from provider-default env vars:
 | Together | `TOGETHER_API_KEY` |
 | Groq | `GROQ_API_KEY` |
 
-Override with `llm.api_key_env` in your config if you have non-default names.
+Override with `--llm-key-env VAR` (or `llm.api_key_env` in config) if you have non-default names. Providers not in the table are resolved by LiteLLM's own per-provider lookup.
+
+**Self-hosted models need no key.** Point `--llm-endpoint` (or `llm.endpoint`) at any OpenAI-compatible server and use `hosted_vllm/<model>` or `openai/<model>`. The provider-default key is never forwarded to a custom endpoint — `openai/` gets a placeholder, `hosted_vllm/` honours `HOSTED_VLLM_API_KEY` if set. Pass `--llm-key-env VAR` to send a specific key. `ollama/<model>` reads `OLLAMA_API_BASE` on its own.
 
 ### Container registry (image distribution for `_runtime` datasets)
 
--- a/docs/reference/BOOTSTRAP.md
+++ b/docs/reference/BOOTSTRAP.md
@@ -64,7 +64,7 @@ The `--llm` flag accepts a LiteLLM-format string. Some examples:
 | Anthropic | `anthropic/<model-name>` |
 | OpenAI | `openai/<model-name>` |
 | Hugging Face Inference Providers | `huggingface/<repo>:<provider>` (e.g. `huggingface/<org>/<model>:together`) |
-| Self-hosted vLLM / Ollama | `openai/<your-model>` with `--llm-endpoint http://...` |
+| Self-hosted vLLM / Ollama / any OpenAI-compatible server | `hosted_vllm/<model>` or `openai/<model>` with `--llm-endpoint http://host:8000/v1` — no API key needed |
 
 Anything LiteLLM supports works, including Bedrock, Vertex, Mistral, and others.
 
@@ -77,11 +77,14 @@ Set the provider's default environment variable; the CLI picks it up automatical
 | Anthropic | `ANTHROPIC_API_KEY` |
 | OpenAI | `OPENAI_API_KEY` |
 | Hugging Face | `HF_TOKEN` |
-| Other | override via `--llm-key-env` or YAML config |
+| Self-hosted (`--llm-endpoint`) | none — the provider-default key is never sent to a custom endpoint. If your server enforces one: `HOSTED_VLLM_API_KEY` for `hosted_vllm/`, or `--llm-key-env VAR` for any route |
+| Other | override via `--llm-key-env` or YAML config; otherwise LiteLLM's own per-provider lookup applies (Bedrock, Vertex, OpenRouter, …) |
 
 ### Cost guardrail
 
 Every bootstrap run is bounded by `max_llm_spend_usd` (default `$5.0`). When the running cost reaches the cap, the agent loop aborts cleanly. Set lower for tighter control:
+
+> Self-hosted models are not in LiteLLM's price table, so their calls count as `$0` and the cap never trips. `max_iterations` / `max_seconds` remain the bounds there.
 
 ```bash
 repo2rlenv generate ... --bootstrap-opt max_llm_spend_usd=1.0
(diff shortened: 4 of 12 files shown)


KNambiarDJsc (commented):
I reviewed and tested b197730:

- **CI-equivalent run on Linux:** ruff is clean, and the full suite passes on py3.12 and py3.14 (752 passed).
- **Key and endpoint handling.** I stubbed `litellm` and compared what reaches `litellm.completion()` on `main` and on this PR:

| # | Scenario | `main` | this PR |
|---|---|---|---|
| 1 | `hosted_vllm/…` + `--llm-endpoint http://127.0.0.1:8000/v1`, no keys | `RuntimeError: no API key resolved` | no `api_key` passed, `api_base` set |
| 2 | `openai/gpt-4o-mini` + endpoint `https://openrouter.ai/api/v1`, `OPENAI_API_KEY` set | `api_key=<OPENAI_API_KEY>` | `api_key='EMPTY'` |
| 3 | `anthropic/…` + a gateway endpoint, `ANTHROPIC_API_KEY` set | `api_key=<ANTHROPIC_API_KEY>` | `api_key='EMPTY'` |
| 4 | row 3 + `api_key_env=ANTHROPIC_API_KEY` | real key | real key |
| 5 | `openai/…` with no endpoint, `api_key_env=MY_GATEWAY_KEY` unset, `OPENAI_API_KEY` set | silently uses `OPENAI_API_KEY` | `RuntimeError: … $MY_GATEWAY_KEY is unset` |
| 6 | `huggingface/…` with no endpoint, `HF_TOKEN` set | `HF_TOKEN` + HF router | unchanged |

- **Provider preflight.** With the real LiteLLM, `check_provider` accepts `hosted_vllm/`, `ollama/`, `openai/` and `huggingface/`, and rejects `antropic/…` before any work starts.

The self-hosted fix works (row 1). Not sending provider keys to an arbitrary endpoint is a good default.

One thing I'd flag before merge: **rows 2, 3 and 5 are silent behavior changes for existing `--config` users**, not just for the new flags.
- **Rows 2 and 3:** some configs point `llm.endpoint` at an authenticated OpenAI- or Anthropic-compatible gateway and rely on the provider-default key. After this PR, those calls send `EMPTY` and will fail with a 401 that doesn't mention the cause.
- **Row 5:** an unset `api_key_env` no longer falls back to the provider default.

Two cheap mitigations:

1. **A runtime warning.** When `endpoint` is set, `api_key_env` isn't, and the provider-default variable is non-empty, log something like: "`$OPENAI_API_KEY` is set but not sent to https://openrouter.ai/api/v1; pass `--llm-key-env OPENAI_API_KEY` to send it."
2. **A release-notes callout.** Mention both changes in `docs/release_notes/HISTORY.md`, since they change how existing config files behave.

adithya-s-k (approved):

adithya-s-k:
thanks @surajsharan, merged! glad we have self-hosted LLM support wired through the CLI now. appreciate the live vLLM testing too.

surajsharan:
> I reviewed and tested [b197730](https://github.com/huggingface/Repo2RLEnv/commit/b1977300a753c5c09daa39a5fce101f3a617fa01):
> 
> * **CI-equivalent run on Linux:** ruff is clean, and the full suite passes on py3.12 and py3.14 (752 passed).
> * **Key and endpoint handling.** I stubbed `litellm` and compared what reaches `litellm.completion()` on `main` and on this PR:
> 
> #	Scenario	`main`	this PR
> 1	`hosted_vllm/…` + `--llm-endpoint http://127.0.0.1:8000/v1`, no keys	`RuntimeError: no API key resolved`	no `api_key` passed, `api_base` set
> 2	`openai/gpt-4o-mini` + endpoint `https://openrouter.ai/api/v1`, `OPENAI_API_KEY` set	`api_key=<OPENAI_API_KEY>`	`api_key='EMPTY'`
> 3	`anthropic/…` + a gateway endpoint, `ANTHROPIC_API_KEY` set	`api_key=<ANTHROPIC_API_KEY>`	`api_key='EMPTY'`
> 4	row 3 + `api_key_env=ANTHROPIC_API_KEY`	real key	real key
> 5	`openai/…` with no endpoint, `api_key_env=MY_GATEWAY_KEY` unset, `OPENAI_API_KEY` set	silently uses `OPENAI_API_KEY`	`RuntimeError: … $MY_GATEWAY_KEY is unset`
> 6	`huggingface/…` with no endpoint, `HF_TOKEN` set	`HF_TOKEN` + HF router	unchanged
> * **Provider preflight.** With the real LiteLLM, `check_provider` accepts `hosted_vllm/`, `ollama/`, `openai/` and `huggingface/`, and rejects `antropic/…` before any work starts.
> 
> The self-hosted fix works (row 1). Not sending provider keys to an arbitrary endpoint is a good default.
> 
> One thing I'd flag before merge: **rows 2, 3 and 5 are silent behavior changes for existing `--config` users**, not just for the new flags.
> 
> * **Rows 2 and 3:** some configs point `llm.endpoint` at an authenticated OpenAI- or Anthropic-compatible gateway and rely on the provider-default key. After this PR, those calls send `EMPTY` and will fail with a 401 that doesn't mention the cause.
> * **Row 5:** an unset `api_key_env` no longer falls back to the provider default.
> 
> Two cheap mitigations:
> 
> 1. **A runtime warning.** When `endpoint` is set, `api_key_env` isn't, and the provider-default variable is non-empty, log something like: "`$OPENAI_API_KEY` is set but not sent to https://openrouter.ai/api/v1; pass `--llm-key-env OPENAI_API_KEY` to send it."
> 2. **A release-notes callout.** Mention both changes in `docs/release_notes/HISTORY.md`, since they change how existing config files behave.

Thanks for the thorough re-test, the table is exactly the comparison I wanted a second pair of eyes on. Rows 2, 3 and 5 were the deliberate trade-off: sending a provider's default key to an arbitrary endpoint felt like the wrong default.

Both mitigations went in before the merge via c6768c5: when a provider-default key is set but withheld from a custom endpoint, the CLI now logs which variable it is and points at `--llm-key-env` / `llm.api_key_env`, and HISTORY.md has the callout for the config-file behaviour change (including row 5). So the 401-with-no-cause case shouldn't happen anymore.

surajsharan:
> thanks @surajsharan, merged! glad we have self-hosted LLM support wired through the CLI now. appreciate the live vLLM testing too.

Thanks for merging and for folding in the warning and release note.

PR #107, opened 2026-09-14
validate: add --deep and --oracle task-aware checks

### LINKED ISSUE #100: Add task-aware asset and reproducibility checks to validate
### Description

`src/repo2rlenv/cli.py:315-356`, `cmd_validate()`, currently parses `task.toml`, checks for `version`, `task` and `task.name` and warns if Repo2RLEnv metadata is absent. It does not validate the task's assets or the contents of its reproducibility configuration.

A task declaring a test-execution reward can therefore be reported as valid even when its instruction, tests, solution and environment are all absent.

The CLI currently advertises a fast structural/TOML check. This proposal extends that capability into an optional deeper preflight check; it should not assume that every Harbor task follows one fixed asset layout.

### Reproduction

Create a directory containing only this `task.toml`:

```toml
version = "1.0"

[task]
name = "review/broken"

[metadata.repo2env]
reward_kinds = ["test_execution"]

[metadata.repo2env.reproducibility]
mode = "nonsense"
```

Run:

```bash
repo2rlenv validate /path/to/directory
```

Actual behavior: validation reports success and returns exit code `0`, despite the invalid reproducibility mode and absence of all task assets. This was reproduced locally through `cmd_validate()`.

### Proposed behavior

Add a documented deep-validation mode, such as `--deep`, that checks the assets required by the declared task format, pipeline and execution configuration and reports actionable paths and errors.

Useful checks include:

- Required instruction content exists and is nonempty for the applicable task format.
- Tasks using a verifier have the required test entry point, accounting for supported environment and task configurations.
- Known Repo2RLEnv verifier dependencies are present and parse correctly. For example, graded `pr_runtime` tasks need their emitted verifier and valid F2P/P2P JSON lists.
- Execution tasks have a usable environment definition under their selected configuration, rather than universally requiring one particular filename.
- Repo2RLEnv reproducibility metadata uses supported modes and satisfies each mode's field requirements, while preserving supported legacy/non-Repo2RLEnv behavior.
- When validating a Repo2RLEnv oracle, the expected solution script and patch exist and are usable. Make oracle validation explicit rather than requiring solutions for every Harbor task.

An optional patch-applicability check could run `git apply --check` against the declared base commit in an isolated checkout. It must use the correct source revision; applying against an unrelated working tree cannot establish task validity.

### Avoid introducing false-positive validation failures

- Harbor solutions are optional; `solution/patch.diff` is a Repo2RLEnv convention, not a universal Harbor requirement.
- A shell script invoked with `bash script.sh` does not require the executable bit. Do not reject otherwise usable tasks solely because scripts are not mode `0755`; check permissions according to the actual invocation requirements.
- Repo2RLEnv supports lightweight `pr_diff` output without environment/tests assets. Deep validation should distinguish this supported output from incomplete runnable tasks.
- Arbitrary shell dependencies cannot be fully established with simple string matching. Start with known emitted assets and clearly state the limits of static validation.

See [Harbor's task documentation](https://www.harborframework.com/docs/tasks) for optional solutions and verifier invocation behavior.

### Impact

Task-aware validation would catch missing files and malformed metadata before expensive environment builds or evaluations, while avoiding rejection of legitimate task variants. Static validation would complement, rather than replace, actual build and oracle verification.

KNambiarDJsc:
## Summary

- Adds `repo2rlenv validate --deep`: task-aware static checks for each task's assets and `[metadata.repo2env]`, reported with task-relative paths. Covers a blank or missing `instruction.md`; runnable tasks with no test entry point or environment definition; graded `pr_runtime` / `commit_runtime` / `cve_patches` tasks whose `verifier.py`, `f2p.json` or `p2p.json` don't parse; `code_instruct` / `equivalence_tests` test files; and `reproducibility` metadata. That last check covers mode ∈ `registry | inline_dockerfile | local_only` plus the per-mode fields.
- Adds `--oracle`, which implies `--deep` and also requires a usable `solution/patch.diff` + solve script. Oracle checks are opt-in, as the issue asks, and the default `validate` pass is unchanged.
- Rules mirror Harbor and our emitter so legitimate variants don't fail:
  - Environment: `docker_image` / `Dockerfile` / `docker-compose.yaml` all count, per Harbor's `has_agent_environment_definition`.
  - Scripts: `.bat` for `os = "windows"`, and no executable-bit check.
  - Output shapes: text-only `pr_diff` output is valid.
  - Legacy metadata: `diff_similarity_multi_component`, still in the published pr_diff reference dataset, is read as `diff_similarity`.

Checks and limits are documented in `docs/reference/SPEC.md#deep-validation`.

Running the issue's reproduction:

```
$ repo2rlenv validate ./broken --deep
✗ review/broken: instruction.md: missing
✗ review/broken: environment: no environment definition: add environment/Dockerfile or environment/docker-compose.yaml, or set [environment].docker_image
✗ review/broken: tests/test.sh: missing
✗ review/broken: task.toml: [metadata.repo2env.reproducibility].mode must be one of inline_dockerfile, local_only, registry; got 'nonsense'
✗ 1/1 tasks failed          # exit 1 (plain `validate` still exits 0)
```

It also escapes Rich markup in `validate` output. Before this, TOML table names like `[metadata.repo2env]` and `[task]` were swallowed as style tags.

## Test plan

- [x] `uv run pytest -q`: 769 passed on Linux py3.12 and py3.14, including the 54 new tests in `tests/test_validation.py`
- [x] `uv run ruff check .` + `uv run ruff format --check .`
- [x] `uv build`
- [x] Fixtures are built with the real emitter + pipeline builders (lite `pr_diff`, `pr_diff` with env, graded `pr_runtime`, `code_instruct`), and each asserts zero findings under `--oracle`
- [x] `validate --oracle` on every published reference dataset: `repo2rlenv-pr-diff` (181), `-pr-runtime` (100), `-commit-runtime` (100), `-code-instruct` (100), `-cve-patches` (19). All 500 tasks came back with 0 errors and 0 warnings.

## Out of scope

- The `git apply --check` patch-applicability check at `base_commit`. It needs an isolated checkout of the source repo, so it's a follow-up; `harbor run --agent oracle` remains the ground truth.
- Per-step checks for multi-step `[[steps]]` Harbor tasks. These skip layout checks with a warning, since Repo2RLEnv doesn't emit them.
- Deep checks on the default `validate` path. That stays fast and keeps its current exit codes.

Closes #100


adithya-s-k (approved):

adithya-s-k:
thanks @KNambiarDJsc, merged! the deeper task checks and validation against the published datasets were really helpful.

Issue #108, opened 2026-09-14
Windows portability: local repo paths, backslash relative paths, and encoding crashes


KNambiarDJsc:
### Description

Running Repo2RLEnv on Windows hits several portability bugs that CI doesn't catch, since it is Ubuntu-only. On a clean clone of `main`, `uv run pytest` fails 7 tests on Windows; the same suite passes under WSL Ubuntu. Every item below was reproduced against `main` @ 1a8bad5.

Is Windows meant to be supported, for generation or at least for contributors running the suite? I'm happy to send PRs for whichever items are in scope. Most are one-line fixes.

**Environment:** Windows 11 Pro (10.0.26200), Python 3.12.11 via uv, repo2rlenv 0.8.8, pytest 9.0.3, Git for Windows.

---

### 1. Windows local paths are rejected or misrouted as GitHub repos

`RepoSpec.normalize_url` only treats a value as local if it starts with `/`, `~`, `./`, `../` or `file://`:

https://github.com/huggingface/Repo2RLEnv/blob/1a8bad5774e062d0d73a7d79f9d6a95540801c83/src/repo2rlenv/spec/input.py#L31-L48

`RepoSpec(url=...)` for a local git repo at `C:\tmp\r2e\origin`:

| `--repo` value | Result |
|---|---|
| `C:\tmp\r2e\origin` | `ValidationError: repo url must be 'owner/name', a full git URL, or a local path` |
| `C:/tmp/r2e/origin` | **silently** `https://github.com/C:/tmp/r2e/origin`, `source_kind=github` |
| `.\origin` | `ValidationError` |
| `file:///C:/tmp/r2e/origin` (what `Path.as_uri()` returns) | `file://C:tmp\r2e\origin`, a drive-relative path, so the clone fails |
| `./origin` or `file://C:\tmp\r2e\origin` | works; the clone succeeds |

**Expected:** drive-letter and backslash paths are classified as local. `file:///C:/...` resolves to `C:\...`, e.g. via `urllib.request.url2pathname`.

**Failing tests:**
- `tests/test_sources.py::test_local_path_canonicalized_to_file_url`
- `tests/test_sources.py::test_resolve_token_local_is_none`
- `tests/test_sources.py::test_local_repo_clones`

### 2. Repo-relative paths use backslashes, which changes task IDs and breaks reference URLs

Both file walkers build `rel = str(path.relative_to(clone_dir))`, which on Windows gives `src\a.py`:

[...]

PR #110, opened 2026-09-14
tests: skip POSIX exec-bit checks on Windows

KNambiarDJsc:
## Summary

- `test_solve_sh_emitted_and_executable` and `test_writes_environment_and_test_script_when_provided` assert `st_mode & 0o111` after the emitter's `chmod(0o755)`. Windows has no POSIX exec bits, so `chmod` does nothing there and both tests fail on every clean clone. This is item 5 of #108.
- Adds a small `_assert_executable()` helper that checks the mode only when `os.name != "nt"`. Every other assertion in both tests still runs on all platforms.

## Test plan

- [x] Windows 11 / py3.12: `tests/test_emitter.py`, 12 passed (was 10 passed, 2 failed). Full-suite failures drop from 7 to 5; the remaining 5 are items 1–2 of #108.
- [x] Linux (WSL Ubuntu, py3.12): full suite, 715 passed
- [x] Linux check that the guard keeps coverage: with the emitter's `chmod(0o755)` calls removed, both tests fail with mode `0644`, so CI still enforces the exec bit
- [x] `uv run ruff check .` + `uv run ruff format --check .`

## Out of scope

- Items 1–4 of #108: Windows local paths, backslash relative paths, console and file encoding. These are left until it's clear whether Windows support is in scope.

Refs #108


adithya-s-k (approved):

adithya-s-k:
thanks @KNambiarDJsc, merged! this keeps the POSIX checks intact while fixing the Windows test failures.

PR #111, opened 2026-09-14
spec: accept Windows local repo paths

KNambiarDJsc:
## Summary

- On Windows, `RepoSpec` now treats drive-letter (`C:\repo`, `C:/repo`), UNC (`\\server\share\repo`) and backslash (`.\repo`) paths as local checkouts. Before this change they were rejected, or in the case of `C:/repo` silently expanded to `https://github.com/C:/repo`. This is item 1 of #108.
- `file:///C:/repo`, which is what `Path.as_uri()` returns, now resolves to `C:\repo`. It used to resolve to the drive-relative `C:repo` and fail to clone.
- The canonical output (`file://<native abspath>`) is unchanged, so `git clone` and existing bootstrap cache keys are unaffected, and POSIX classification is untouched. The rules live in two small pure helpers that take the platform as a flag, so Linux CI exercises the Windows branch.

On Windows, every input form now normalizes to the same URL and clones successfully:

| `--repo` | before | after |
|---|---|---|
| `C:\tmp\r2e\origin` | `ValidationError` | `file://C:\tmp\r2e\origin` |
| `C:/tmp/r2e/origin` | `https://github.com/C:/tmp/r2e/origin` | `file://C:\tmp\r2e\origin` |
| `.\origin` | `ValidationError` | `file://C:\tmp\r2e\origin` |
| `file:///C:/tmp/r2e/origin` | `file://C:tmp\r2e\origin`, and the clone fails | `file://C:\tmp\r2e\origin` |
| `./origin`, `file://C:\tmp\r2e\origin` | worked | unchanged |

## Test plan

- [x] New tests in `tests/test_sources.py` run on every OS:
  - Windows path forms are local with `windows=True` and not with `windows=False`.
  - Remotes are never classified as local paths.
  - `file:///C:/...` is parsed correctly.
  - A native path, `Path.as_uri()` and the canonical URL all normalize to the same value.
- [x] Windows 11 / py3.12: the 3 previously failing `test_sources.py` tests now pass. A real `git clone` succeeds for all four Windows forms in the table.
- [x] Linux (WSL Ubuntu): 726 passed on py3.12 and on py3.14. `ruff check` and `ruff format --check` are clean.
- [x] Combined with #110 and the companion PR for item 2, the full suite passes on Windows (728 passed, 0 failed).

## Out of scope

- Percent-encoded `file://` URLs (e.g. `%20` from `as_uri()` for paths with spaces). These weren't decoded before on any OS either.
- Items 3–4 of #108 (console and file encodings).

Docs that list accepted `--repo` forms (`README.md`, `docs/pipelines/README.md`, `docs/reference/AUTH.md`, `CLAUDE.md`) now mention the Windows forms.

Refs #108


adithya-s-k (approved):

adithya-s-k:
thanks @KNambiarDJsc, merged! appreciate the Windows path fixes and the repros.

PR #112, opened 2026-09-14
pipelines: emit POSIX repo-relative paths on every OS

KNambiarDJsc:
## Summary

- `walk_repo` (`equivalence_tests`) and `sample_seed` (`code_instruct`) built repo-relative paths with `str(path.relative_to(clone_dir))`, which gives `src\a.py` on Windows. They now use `.as_posix()`. This is item 2 of #108.
- That string is hashed into `task_id` and the `test_r2e_*.py` filename. It is also written into the `reference` URL (`.../blob/<ref>/src\a.py#L1-L2`), `seed_path` / `source_path` metadata, and the LLM prompt. Before this change, the same function got a **different task ID depending on the OS that generated it**. POSIX output is byte-for-byte unchanged.
- `list_source_files` gets the same change for consistency. `fnmatch` normcases separators on Windows, so exclusion already worked there. Its test also no longer builds backslash paths itself.

## Test plan

- [x] `test_sample_seed_returns_substantive_window` is now parametrized over `real.py` and `pkg/sub/real.py`.
- [x] Windows 11 / py3.12, with the source fix stashed: `test_sample_seed_returns_substantive_window[pkg/sub/real.py]` and `test_walk_repo_respects_globs` fail. With the fix, all 44 tests in both files pass.
- [x] Linux (WSL Ubuntu): 716 passed on py3.12 and on py3.14. `ruff check` and `ruff format --check` are clean.
- [x] Combined with #110 and the companion PR for item 1, the full suite passes on Windows (728 passed, 0 failed).

## Out of scope

- Datasets previously generated on Windows keep their backslash-derived task IDs. Regenerating them yields the same IDs as Linux.
- Items 3–4 of #108 (console and file encodings).

Refs #108


adithya-s-k (approved):

adithya-s-k:
thanks @KNambiarDJsc, merged! keeping task IDs and paths consistent across platforms is important.

PR #114, opened 2026-09-15
cli: don't crash printing to a non-UTF-8 stdout

KNambiarDJsc:
## Summary

- `cli.main()` now calls a new `ui.ensure_utf8_output()`, which switches stdout and stderr to UTF-8 when their encoding can't represent the UI glyphs. This is item 3 of #108. On Windows, pipes and redirected files use the ANSI code page (cp1252), so the first `✓`/`✗`/`⚠` raised `UnicodeEncodeError` and commands exited 1 even when they succeeded. UTF-8 streams and an explicit `PYTHONIOENCODING` are left untouched.
- `push_to_hub` now reads the source `manifest.json` as UTF-8, matching how it already writes it. This covers the `hub.py` part of item 4. On Windows, a non-ASCII enriched manifest failed to decode, was treated as not enriched, and its `validation` block was silently replaced by the regenerated minimal manifest.

Before and after on Windows, running `repo2rlenv validate` on a valid dataset:

| stdout | `main` | this PR |
|---|---|---|
| piped | exit 1, `UnicodeEncodeError: 'charmap' codec can't encode character '✓'` | exit 0 |
| redirected (`> out.txt`) | exit 1, same error | exit 0; `out.txt` is valid UTF-8 |

## Test plan

- [x] New `tests/test_ui_console.py`, which runs on every OS:
  - a cp1252 stdout or stderr is switched to UTF-8
  - `console.success` on a cp1252 stream writes UTF-8 (without the fix it raises `UnicodeEncodeError`)
  - `PYTHONIOENCODING` is respected
  - UTF-8 streams and streams without `reconfigure()` are left alone
- [x] `test_push_preserves_enriched_manifest` is now parametrized with a non-ASCII title. It fails on Windows without the `hub.py` change and passes with it.
- [x] Windows 11: the #108 reproduction (piped and redirected) exits 0.
- [x] Linux (WSL Ubuntu): 721 passed on py3.12 and on py3.14. `ruff check` and `ruff format --check` are clean.

## Out of scope

- Decoding subprocess output. That's a separate PR.
- `validate`'s own `task.toml` read, which is already covered by #107.
- `bootstrap/cache.py` and `auth.py`, listed under item 4. They read ASCII-only files (the cache JSON uses `json.dumps`' default `ensure_ascii=True`, and HF tokens are ASCII), so they're not affected.

Refs #108


adithya-s-k (approved):

adithya-s-k:
thanks @KNambiarDJsc, merged! good to have piped output and non-ASCII manifests handled properly on Windows.

PR #115, opened 2026-09-15
decode subprocess output as UTF-8 on every OS

KNambiarDJsc:
## Summary

- All 20 text-mode `subprocess` calls now pass `encoding="utf-8"`. They cover gh, git, docker, harbor and the registry CLIs.
  - With `text=True` alone, Python decodes output with the locale encoding: UTF-8 on Linux and macOS, but cp1252 on Windows. On Windows, non-ASCII tool output was silently lost or corrupted.
  - The change makes Windows decode exactly as Linux already does. Error handling is unchanged (strict).
- Adds `tests/test_subprocess_encoding.py`, a static check that fails if a new text-mode subprocess call omits `encoding`. Linux CI can't observe the bug itself, so this keeps new calls explicit.
- I found this while working on #108 and am adding it there as item 6.

Reproduced on Windows against live data:

| Call | `main` | this PR |
|---|---|---|
| `github._run_gh`: `gh api` for issue #108's body (contains `Árvíztűrő`) | **empty**: `stdout=None`, because the `UnicodeDecodeError` is swallowed in subprocess's reader thread | 6408 chars, exact |
| `github.fetch_pr_diff` for PR #107 | **garbled**: `├──` becomes `â”œâ”€â”€`, and that text is written to `solution/patch.diff` | byte-exact |
| `git_local._run_git`: `git log` and `git show` of a commit whose subject contains `修复` | **empty** | exact |

## Test plan

- [x] `tests/test_subprocess_encoding.py` fails on `main`, listing all 20 call sites, and passes here.
- [x] Windows 11 / py3.12: the live reproduction above passes.
- [x] Windows 11 / py3.12: `tests/test_e2e_public.py::test_e2e_public_trl` passes; it's a real `pr_diff` run against huggingface/trl. On `main` it fails with `json.loads(None)` because a recent trl PR contains such a character; that isn't the `diff_too_small` flake. The test's own `patch.diff` read also needed `encoding="utf-8"`.
- [x] Linux (WSL Ubuntu): 716 passed on py3.12 and on py3.14. Ruff is clean.
- [x] Merged together with #107, #110, #111, #112 and the console-encoding PR, the full suite passes on Windows without `PYTHONUTF8`: 789 passed, 0 failed.

## Out of scope

- The `errors=` policy for tool output that isn't valid UTF-8. It stays strict, the same as Linux today.

Refs #108


adithya-s-k (approved):

adithya-s-k:
thanks @KNambiarDJsc, merged! appreciate you tracking down the subprocess encoding issue and adding coverage for it.

PR #129, opened 2026-09-15
fix: restore Windows CLI and preserve controller locks

### LINKED ISSUE #128: v0.9.0 / main: CLI is completely broken on Windows (ModuleNotFoundError: fcntl) — includes --version
### Severity: release-blocking — `pip install repo2rlenv` is completely non-functional on Windows right now

Every `repo2rlenv` CLI invocation crashes on Windows before argument parsing even completes, including `--version`. This is not on a branch — it's in the **current PyPI release** (confirmed below), so anyone on Windows who installs the package today gets a tool that cannot run a single command.

### Reproduction

```bash
# Fresh venv, real PyPI install — no git, no fork, no local state
uv venv && uv pip install repo2rlenv
repo2rlenv --version
```

```
ModuleNotFoundError: No module named 'fcntl'
```

Same result from a from-scratch `git clone https://github.com/huggingface/Repo2RLEnv.git` + `uv sync --group dev` + `uv run repo2rlenv --version`, and from the `main` branch tip directly (`23616db`) — so it isn't specific to the PyPI packaging step either.

**Environment:** Windows 11 Pro, Python 3.12.11.

### Root cause

`cli.py`'s dispatcher imports several new subsystems unconditionally, before routing to any subcommand — so this runs on *every* invocation, `--version` included:

```python
# main() -> _dispatch()
from repo2rlenv.quality.loop.cli import add_quality_parser   # cli.py:1036
...
from repo2rlenv.tasksmith.cli import add_parser as add_tasksmith_parser
```

Four files reachable from that chain do `import fcntl` at module level — `fcntl` is POSIX-only and does not exist in Windows' Python stdlib at all, so no amount of `pip install`-ing fixes it:

- `src/repo2rlenv/quality/loop/runner.py:5`
- `src/repo2rlenv/tasksmith/runner.py:6`
- `src/repo2rlenv/tasksmith/batch.py:6`
- `src/repo2rlenv/pipelines/recipes/history/worker.py:6`

In each case it's used for the same thing — advisory, non-blocking file locking to stop two concurrent controllers/loops from running against the same output directory:

```python
with (self.directory / ".lock").open("a") as lock:
    fcntl.flock(lock, fcntl.LOCK_EX | fcntl.LOCK_NB)
```

Trace confirming the exact import chain (`_dispatch` → `quality.loop.cli` → `quality.loop.runner` → `import fcntl` → `ModuleNotFoundError`) is in the repro above; happy to paste the full `builtins.__import__` trace if useful.

### Timeline

- `be393f9` (#109, "feat: consolidate Tasksmith and owned Harbor generation recipes") merged **2026-09-15T11:03:56Z** — 468 files, +59,156/-187, **0 reviews**, self-merged by its author.
- All CI checks on that PR were green (`Tests (py3.12/3.13/3.14)`, `Lint`, `Build`, etc.) — CI runs Ubuntu only, and `fcntl` exists on Linux, so nothing in the matrix could have caught this.
- `v0.9.0` was tagged and released to PyPI **2026-09-15T11:53:16Z**, ~50 minutes later, from a commit that includes `be393f9`. Confirmed via `git merge-base --is-ancestor be393f9 v0.9.0` and by installing `repo2rlenv==0.9.0` from PyPI directly (repro above).

### Proposed fix

Same shape at all four call sites — guard the import and the lock acquisition by platform, since `fcntl.flock` has no drop-in Windows equivalent in the stdlib (`msvcrt.locking` is the nearest analog but has a different API; the `filelock` package, already a common transitive dep in this kind of stack, is the more portable fix if a third dependency is acceptable):

```python
import sys

if sys.platform != "win32":
    import fcntl

...
with (self.directory / ".lock").open("a") as lock:
    if sys.platform != "win32":
        fcntl.flock(lock, fcntl.LOCK_EX | fcntl.LOCK_NB)
    # else: single-controller-per-directory isn't enforced on Windows yet
```

Separately, and lower-priority than the crash itself: `_dispatch()` eagerly importing every subsystem (`campaigns`, `quality.loop`, `tasksmith`, `pipelines.recipes`) just to register their argument parsers means a bug in *any* of them (not just this one) takes down the entire CLI, including `--version` and `--help`. Lazily importing each subparser's module only when that subcommand is actually invoked (argparse supports this via a thin wrapper, or `add_parser(..., func=lambda args: _lazy_import_and_run(args))`) would contain future failures to the subcommand that has them, rather than the whole binary.

Happy to send a PR for the four-file `fcntl` fix if useful — it's mechanical and doesn't touch the locking semantics on POSIX.

KNambiarDJsc:
## Summary

Windows installations of 0.9.0 crash on every CLI command, including `--version` and `--help`, because parser registration reaches four modules that import POSIX-only `fcntl`.

Use a shared standard-library lock with lazy platform imports: retain `flock` on POSIX and lock byte zero through `msvcrt` on Windows. Controller locks remain nonblocking; the history checkout lock still waits for its owner. Closing the handle releases either lock, and the lock path is retained.

Add fresh-wheel Windows CI for Python 3.12–3.14, including CLI discovery, native UTF-8 task emission/static validation and real subprocess lock contention. Publication depends on the same Windows gate. Document the current host support scope. Thanks to @KNambiarDJsc for the report, initial fix and import regression test.

## Test plan

- [x] Full macOS all-extras suite: 1,874 passed, 4 optional live checks skipped.
- [x] Ruff lint and formatting.
- [x] Fresh base-only wheel: CLI entrypoint, help, all 14 recipe descriptions, structured errors and native task validation.
- [x] Real subprocess tests: conflicting locks, blocking acquisition, normal/abrupt owner exit and POSIX compatibility with earlier releases.
- [x] Native Windows wheel checks on Python 3.12, 3.13 and 3.14.
- [x] Linux CI suite, runtime SDK contracts and strict docs build.

## Out of scope

Full native Windows Tasksmith, research-recipe and quality-controller execution is tracked in #130. Use Linux, macOS or WSL for those controllers; choosing a remote sandbox does not remove host artifact/process requirements. No generated datasets need rewriting for this CLI/locking fix.

Closes #128


adithya-s-k:
thanks for catching this and putting up the fix. i added real Windows locks so parallel controllers cannot write the same run, plus Windows wheel checks in CI and the release gate. local focused tests are passing; waiting on the native Windows jobs before merging. broader controller portability will be tracked separately.

adithya-s-k (approved):
verified the final changes: locks remain exclusive on both platforms, the shared checkout still waits correctly, and all CI checks pass including Windows 3.12–3.14. thanks for the report and initial fix. the wider portability work is tracked in #130.

KNambiarDJsc:
Nice — `msvcrt.locking` + a manual retry loop for the blocking case is the right call, and the subprocess-based tests are exactly the hard cases that matter here (real cross-process contention, not mocked).

Pulled 640fbc into a clean worktree and ran it on a separate Windows 11 machine, independent of CI:
- `repo2rlenv --version` / `--help`: exit 0, full subcommand list ✓
- `tests/test_locking.py`: 4 passed, 1 skipped (the POSIX-interop test, correctly platform-gated) — including the blocking-wait/acquire handoff and lock-released-on-process-death cases
- `tests/test_posix_only_imports.py`, `tests/check_cli_install.py`: pass

All green independently of the CI run. Looks ready from my side.

adithya-s-k:
merged and released in 0.9.1. Windows CI and the fresh PyPI install checks passed. thanks for catching this and getting the fix started!

Issue #130, opened 2026-09-15
Track native Windows controller and artifact portability


adithya-s-k:
The CLI startup regression in #128 is being fixed in #129, with real Windows process locks and fresh-wheel checks on Python 3.12–3.14. Full native Windows controller support needs separate work.

The contributor reported 249 failures and 97 errors in the all-extras Windows suite. Those counts are their report, not independently reproduced here. Local code inspection identified these concrete gaps:

- Bundle identities require POSIX 0644/0755 modes. Windows chmod cannot represent those modes; preserve meaningful, stable permissions in manifests and exported archives without weakening identity checks.
- Tasksmith controller snapshots use POSIX permissions, and local author cleanup uses process groups / killpg. The OpenCode adapter explicitly requires Linux or macOS for cleanup. Add a supported Windows lifecycle or fail before allocating paid work.
- Audit host-side text I/O and subprocess encodings for UTF-8 and line endings. Keep Linux sandbox behavior separate from host portability.

For now, use Linux, macOS or WSL for Tasksmith, research-recipe generation and quality-loop execution. A remote Daytona/Modal worker does not remove host-side requirements. Basic installed CLI, discovery, native emission and static validation have a dedicated Windows CI gate.

Acceptance: reproducible categorized Windows failures; targeted fixes with native Windows tests; unchanged bundle identity, controller exclusivity and budget protections; a documented supported-command matrix. This is not a reason to skip or relax artifact integrity checks.

KNambiarDJsc:
Independently reproduced, on a separate machine from where I first reported the counts, on the current `main` tip (`022d1bd`, `repo2rlenv` 0.9.1 — after #129, so unrelated to the startup crash):

```
uv sync --group dev --all-extras
uv run --all-extras pytest -q
```

| | Linux (WSL Ubuntu) | Windows 11 |
|---|---|---|
| Result | 1866 passed, 0 failed | **1525 passed, 249 failed, 97 errors, 7 skipped** |

Same 249/97 as before — nothing excluded, and I confirmed neither number is an artifact of my earlier `--group dev`-only (non-`--all-extras`) setup.

I traced the actual error text behind every failure/error rather than just counting test names, so here's the categorized, reproducible breakdown your acceptance criteria asks for:

**1. The `0o644`/`0o755` bundle-mode check — the large majority of both counts.**
```
src/repo2rlenv/emitter/bundle.py:158
```
```python
mode = item.stat().st_mode & 0o7777
[...]

PR #136, opened 2026-09-16
fix: Windows portability for os.chown and git-apply CRLF hermeticity

KNambiarDJsc:
## Summary

Two of the mechanical root causes from the #130 Windows-portability triage (#2 and #4):

**#2 — `os.chown` doesn't exist on Windows.** `grade.py`'s two call sites now guard on `sys.platform`. Both only ever run for real inside the Linux verifier container regardless of host OS (every other path in that module is an absolute container path like `/workspace`, `/tests`), so the guard changes no production behavior — it only lets the tests that call `validate_submission()` directly run on a Windows host. Four test files `monkeypatch.setattr("os.chown", ...)` to neutralize the call during unit tests; switched to `raising=False` so the patch itself doesn't `AttributeError` on a platform where the attribute never existed.

**#4 — `git apply --reverse` isn't hermetic against the caller's git config.** `reverse_source()` and `reverse_crlf_patch()` both run `git apply` against plain directories (not git repositories), so without an explicit override they silently inherit the *caller's global* git config. On a machine with `core.autocrlf=true` — the Git-for-Windows installer default — `git apply` was rewriting LF to CRLF while applying, corrupting the exact-byte reversal both functions promise and breaking the dedicated `reverse_crlf_patch()` fallback, which assumes the applied bytes are exactly what it constructed and verifies that via git blob hashes. Both invocations now pass `-c core.autocrlf=false` explicitly.

While tracking that down I also found two `write_text()` calls doing the same kind of damage independently of `git apply`: the reversed patch file (`reverse_source()`) and the private verifier test source (`construct()`) are both read back later as raw bytes, but were written with `write_text()`, whose newline translation rewrites `\n` to the host line separator on the way in. Switched both to `write_bytes()`.

Also skips `test_verifier_editor_backups.py`'s `"fifo"` parametrize case on `win32`: `os.mkfifo` has no Windows equivalent at all, so that one case can't run there (not a permissions issue like the neighboring symlink cases, which are left alone — those need your call on whether Developer Mode is a documented prerequisite, per my earlier comment on #130).

## Test plan

- `uv run ruff check` / `ruff format --check` on all touched files — clean.
- Windows 11, full suite before/after: **249 failed, 97 errors → 242 failed, 74 errors** (net -30). Grepped the full run for any remaining `os.chown` / `os.mkfifo` / CRLF-patch-apply failures — none; the only matches left are my own guard code and the one intentional skip.
- Linux (WSL Ubuntu, `uv sync --group dev --all-extras --frozen` + `uv run --all-extras pytest -q`, matching CI): **1871 passed, 0 failed** — including the `fifo` and symlink cases this skip doesn't touch on that platform.
- The 7 Windows failures remaining in the files this PR touches are all the already-tracked symlink-privilege issue (`WinError 1314`, root cause #3 in #130) — untouched here, pending your decision on documenting Developer Mode as a prerequisite.

## Out of scope

Root cause #1 (`emitter/bundle.py`'s mode-check tied to the bundle's content-addressed integrity hash) and root cause #3 (symlink privilege) — both flagged in my correction comment on #130 as needing your design/product decision, not something to ship unilaterally as a platform guard.

Closes part of #130.


adithya-s-k:
Thanks, tested and merged. Reproduced the autocrlf issue and confirmed the fix. Keeping #130 open for the remaining Windows work.

PR #137, opened 2026-09-17
fix(reward): ignore git extended headers and mode lines in diff norma…

shobhitagnihotri69:
### Summary
Ensures `_normalize_diff` in `repo2rlenv.reward` strips git extended headers (e.g. `new file mode`, `old mode`, `deleted file mode`, `similarity index`, and rename metadata) as well as `\ No newline at end of file` markers.

### Problem
When evaluating agent-generated patches in `calculate_diff_similarity_reward`, if the model output matches the oracle code changes but differs in permission bits, file creation modes, or newline markers, the unstripped metadata lines cause the reward to degrade (e.g., scoring ~0.88 instead of 1.0). In an RL training loop, this introduces false reward penalties on valid solutions.

### Solution
- Added `_GIT_EXTENDED_HEADER_RE` to ignore mode changes and copy/rename metadata lines.
- Added `_NO_NEWLINE_RE` to ignore git newline notice markers.
- Added unit test `test_normalization_ignores_git_extended_headers_and_mode_changes` verifying identical code changes with extended headers score 1.0.

### Verification
- `pytest tests/test_reward.py`: All 6 tests passing.
- `pytest tests/test_grading.py`: All 6 tests passing.
- `ruff check`: All checks passed.


adithya-s-k:
Thanks, tested this and found a regression: an identical mode-only patch (100644 to 100755) scores 0 instead of 1 because normalization removes the whole oracle. Please preserve mode/rename-only changes and add regression tests. Also run ruff format on tests/test_reward.py; the extra blank line currently fails the check.

shobhitagnihotri69:
Thanks for catching that @adithya-s-k! In commit `32b297a`, I updated `_normalize_diff` to distinguish content patches from mode/rename metadata: when no hunks are present, git extended headers and file path headers are preserved so mode-only (e.g. 100644 to 100755) and rename-only changes are never stripped to empty diffs. 

I've also added regression tests covering identical mode patches (scoring 1.0), mode mismatches, rename-only patches, and multi-file diffs, and ran `ruff format` on `tests/test_reward.py`.

adithya-s-k:
Thanks for addressing the mode-only case. The reward regressions pass now; merged.

PR #139, opened 2026-09-17
log_parsers: keep cargo should_panic and doctest results

### LINKED ISSUE #138: cargo parser drops should_panic tests and doctests
### Description

`src/repo2rlenv/log_parsers/cargo_parser.py` (L40-42) and its standalone copy in `src/repo2rlenv/pipelines/_pr_runtime_verifier.py` (L142) read the test name as `\S+`. libtest writes `test {name} - {test_mode} ... ` for `#[should_panic]`, `compile_fail` and `no_run` tests, and doctest names contain spaces ([`write_test_name`](https://github.com/rust-lang/rust/blob/master/library/test/src/formatters/pretty.rs), [`TestDesc::test_mode`](https://github.com/rust-lang/rust/blob/master/library/test/src/types.rs)):

```
test tests::panics_on_zero - should panic ... ok
test src/lib.rs - add (line 30) - compile fail ... ok
test src/lib.rs - add (line 34) - compile ... ok
test src/lib.rs - math::div (line 10) ... ok
```

None of these lines match, so the tests never reach the status map. On a real `cargo test --no-fail-fast` run (rustc 1.98.1) that covers every test mode, main records 6 of 18 results. Every `should panic` test and every doctest is missing.

### Impact

- A fix whose new test is `#[should_panic]` gets no FAIL_TO_PASS test. Validation ends with "no fail-to-pass tests after validation" (`pr_runtime_validate.py` L301), and the candidate is dropped. This affects `pr_runtime`, `commit_runtime` and `cve_patches`.
- Doctests never contribute to F2P or P2P.

### Reproduction

A small crate with `tests/div.rs`:

```rust
#[test]
fn divides() { assert_eq!(scenario::div(6, 3), 2); }

#[test]
#[should_panic(expected = "divide by zero")]
fn rejects_zero() { scenario::div(1, 0); }
```

At base, `div` returns 0 for a zero divisor. The gold patch makes it `panic!("divide by zero")`. Real logs:

```
# pre (base + tests)
test divides ... ok
test rejects_zero - should panic ... FAILED
test src/lib.rs - div (line 3) ... ok

# post (gold patch)
test divides ... ok
test rejects_zero - should panic ... ok
test src/lib.rs - div (line 3) ... ok
```

On main, F2P is `[]` and P2P is `["divides"]`, so the task is dropped.

### A trap in the obvious fix

Allowing spaces in names isn't enough, because doctest names carry `(line N)`. That number moves whenever a patch edits code above the doctest. With names kept verbatim, P2P becomes `["divides", "src/lib.rs - div (line 3)"]`.

A correct agent patch that adds a helper function above `div` produces `test src/lib.rs - div (line 9) ... ok`. Given that real run, the verifier returns `reward=0.5`, `resolved=False`, and lists the doctest as a regression.

### Proposed fix

- Strip the mode suffix (` - should panic`, ` - compile fail`, ` - compile`). The identity is the test path, `tests::panics_on_zero`.
- Key doctests by file and item without the line number (`src/lib.rs - math::div`, or `src/lib.rs` for crate-level docs). When an item has several doctests, the worst status wins, so a line shift can't change the identity.
- Make the same change in both copies, with tests that run against both parsers (the same approach as #91).

I have a PR ready and will link it here.

k21993:
## Summary

- Parse libtest result lines whose names contain a test mode or spaces (`tests::x - should panic`, `src/lib.rs - add (line 30) - compile fail`), in both `log_parsers/cargo_parser.py` and the standalone copy in `_pr_runtime_verifier.py`. On main, these lines don't match, so a fix whose new test is `#[should_panic]` has no FAIL_TO_PASS test and gets dropped.
- Key doctests by file and item without `(line N)`, and let the worst status among an item's doctests win. The line number moves whenever a patch edits code above a doctest, so a verbatim key would score a correct patch that shifts lines as a regression.
- Match libtest's literal ` ... ` separator instead of `\s+\.\.\.\s+`, so names can contain spaces without backtracking on long lines.

The reproduction from #138 uses real `cargo test` output. The reference fix keeps the doctest at `(line 3)`. The agent's patch is equally correct but adds a helper above `div`, which moves the doctest to `(line 9)`.

| Parser | F2P | P2P | Reference fix graded | Agent patch graded |
| --- | --- | --- | --- | --- |
| main | `[]` | `divides` | task dropped | task dropped |
| spaces allowed, names verbatim | `rejects_zero` | `divides`, `src/lib.rs - div (line 3)` | 1.0 | **0.5**, doctest reported as a regression |
| this PR | `rejects_zero` | `divides`, `src/lib.rs - div` | 1.0 | 1.0 |

## Test plan

- [x] `uv run --all-extras pytest -q`: 1912 passed, 6 skipped (live-network tests)
- [x] `uv run ruff check .` and `uv run ruff format --check .`
- [x] New `tests/test_cargo_test_names.py` runs every case against both parsers, the same way `tests/test_pytest_node_ids.py` does. Its fixtures are verbatim `cargo test --no-fail-fast` output from rustc 1.98.1: one crate covering every test mode, plus the pre-fix, reference-fix and agent runs above.
- [x] 25 of the 40 new tests fail on main, and all 40 pass with this change.
- [x] The standalone verifier, run with `python -I`, grades the shifted-doctest agent run at reward 1.0 with `resolved=True`.
- [x] Four 200k-character pathological lines parse in a few milliseconds, with a subprocess timeout guard like the one in #91.

## Out of scope

- Tests with the same name in different crates or test binaries still collide, and the last write wins, as before.
- `cargo nextest`, `--format terse` or JSON output, and ANSI-colored logs.
- Tasks generated before this change keep the verifier baked into them.

Closes #138


adithya-s-k:
Thanks, tested and merged. The should_panic and shifted-doctest cases now work in both parsers. Existing Cargo tasks will need their verifier and test IDs updated together.

PR #140, opened 2026-09-17
tests: fake os.killpg/signal.SIGKILL for Windows in remote-job test

KNambiarDJsc:
## Summary

Continuing the Windows-portability triage from #130. `test_timeout_terminates_process_group_and_its_owned_containers` mocks both `Popen` and `killpg` to exercise `execute()`'s timeout/cleanup state machine without touching a real OS — but `monkeypatch.setattr`'s default `raising=True` requires the target attribute to already exist, and neither `os.killpg` nor `signal.SIGKILL` exists on Windows at all. The test failed before its fake was even installed:

```
AttributeError: <module 'os' (frozen)> has no attribute 'killpg'
```

and, after patching that, a second one:

```
AttributeError: module 'signal' has no attribute 'SIGKILL'. Did you mean: 'SIGILL'?
```

`execute()` only ever runs for real inside a Linux remote worker (`REPO2RLENV_REMOTE_WORKER=1`; `killpg` needs a POSIX process group), so this doesn't touch production code — same reasoning as the `os.chown` guards in #136, just with `raising=False` letting the existing fakes install cleanly and the test verify the real cleanup logic from any host OS.

## Test plan

- [x] `uv run pytest tests/test_remote_job.py -q` — 3 passed (was 1 failed before this change)
- [x] `uv run ruff check` / `ruff format --check` on the touched file
- [x] Full suite still collects and runs cleanly with this change (no new failures introduced; `test_remote_job.py` no longer appears in the Windows failure set)


adithya-s-k:
Thanks, the Windows test stub is fixed without changing the Linux worker behavior. Tests and CI pass; merged.

PR #141, opened 2026-09-17
tmax: decode taxonomy.json and template_prompt.md as UTF-8

KNambiarDJsc:
## Summary

Another item from the #130 Windows-portability triage: `tmax.sampler.template_prompt()` reads two package data files via `importlib.resources.files(...).read_text()` with no explicit encoding, so Python falls back to the locale's preferred encoding — `cp1252` on Windows, not UTF-8. `template_prompt.md`'s prose contains bytes `cp1252` can't decode (a right single quotation mark, byte `0x9d`), so every call crashes:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 2918: character maps to <undefined>
```

This is the same bug class fixed in #115 (locale-default text decoding), but a different call shape — `importlib.resources.files().read_text()` rather than `subprocess(text=True, ...)` — so it was outside that PR's scope and got missed.

Scoped narrowly to the two calls in this function; a repo-wide sweep of every `.read_text()`/`.write_text()` call is a much bigger, separate change and most of those sites only ever round-trip the tool's own generated JSON/TOML (never non-ASCII prose), so they're not exhibiting this failure mode.

## Test plan

- [x] `uv run pytest tests/test_tmax_recipe.py::test_taxonomy_sampling_is_repeatable_and_preserves_legacy_axes -q` — now passes (was failing with `UnicodeDecodeError` on Windows)
- [x] `uv run ruff check` / `ruff format --check` on the touched file


adithya-s-k:
Thanks, explicit UTF-8 is the right fix here. Tests and CI pass; merged.

PR #143, opened 2026-09-17
pr_runtime: run vitest with its verbose reporter and parse it

### LINKED ISSUE #142: pr_runtime can't use vitest repos: --verbose crashes and output doesn't parse
### Description

Repos tested with vitest can't produce `pr_runtime` or `cve_patches` tasks. Three things go wrong in sequence.

**1. The normalized command crashes.** `normalize_test_cmds_for_runtime` (`src/repo2rlenv/pipelines/pr_runtime.py` L675-682) appends `--verbose` to any command containing `jest|mocha|vitest`. Vitest has no such flag:

```
$ npx vitest run --verbose
CACError: Unknown option `--verbose`
```

Reproduced on vitest 3.2.7, 4.1.11 and 5.0.1. No test runs, so pre and post both parse to `{}`.

**2. Vitest output is colored in the sandbox.** Vitest's bundled tinyrainbow enables color unless `NO_COLOR`, `TERM=dumb` or `FORCE_TTY=false` is set, even when output is piped. A non-TTY `docker exec` has no `TERM`, so every line starts with an escape code, and `_JEST_TEST_RE` (`log_parsers/jest_parser.py` L54-56, `_pr_runtime_verifier.py` L177-179) never matches. With an empty environment (`env -i PATH=... HOME=...`), `npx vitest run --reporter=verbose` parses to `{}` on main.

**3. Uncolored lines keep the duration in the name.** The verbose reporter prints ` ✓ math.test.ts > math > adds 1ms`, a bare `1ms` rather than jest's `(1 ms)`. So the test is `math.test.ts > math > adds 1ms` in one run and `... adds 4ms` in the next, and F2P/P2P names stop matching at grading time.

### A trap in the obvious fix

Parsing vitest's *default* reporter isn't safe, even after stripping color. It expands a file only while that file has a failure, and collapses a fully passing file to one line. Real vitest 5.0.1 output for a fix to `div`, where the gold patch leaves an unrelated test in the same file failing:

```
# gold patch: file still has a failure, so tests are listed
   ❯ math (3)
     ✓ adds 1ms
     ✓ rejects division by zero 0ms
     × rounds half up 3ms

# agent patch that also fixes `rounds half up`
 ✓ math.test.ts (3 tests) 2ms
```

F2P would be `rejects division by zero`, and the agent's more complete patch would lose it from the log and score 0.

### Proposed fix

- For vitest commands, drop `--verbose` and append `--reporter=verbose` (unless a `--reporter` is already set). Jest and mocha keep `--verbose`.
- In both parsers, strip ANSI escape codes. When the log has vitest's ` RUN  vX.Y.Z` or ` Test Files` marker, record only verbose lines (`✓ file > suite > test`), dropping vitest's run suffixes (duration, `(retry xN)`, `(repeat xN)`, heap usage, skip note). Keep a project label (`|unit|`) as part of the name, so the same file run in two projects stays distinct.
- Accept jest's padded colored header (` FAIL  file`), so colored jest output parses the same as plain output.

I have a PR ready and will link it here. Mocha has a separate parsing problem, and I'll file it on its own.

k21993:
## Summary

- `normalize_test_cmds_for_runtime` now gives vitest `--reporter=verbose` instead of `--verbose`. Vitest 3-5 reject `--verbose` with `CACError: Unknown option`, so no test ran.
- Both JS parsers strip ANSI escape codes, which vitest emits whenever `TERM` is unset (as in a non-TTY `docker exec`). In a vitest log, they read only the fully qualified verbose lines and drop the run suffixes (` 12ms`, `(retry xN)`, `(repeat xN)`, heap usage, skip note). A project label stays in the name as `|unit| file > suite > test`, whether or not the output is colored.
- Vitest's default reporter is ignored on purpose. It collapses a fully passing file to one line, so a patch that fixed more than the gold patch would lose its F2P tests (details in #142).
- Jest's padded colored header (` FAIL  file`) is accepted, so colored jest output parses the same as plain jest output.

Results on a real vitest 5.0.1 bug fix (the `div` fix from #142, captured with an empty environment):

| | Command run | F2P | Agent patch graded |
| --- | --- | --- | --- |
| main | `npx vitest run --verbose` → `CACError`, no tests | none | task dropped |
| main parser, even with the right flag | colored verbose output → `{}` | none | task dropped |
| this PR | `npx vitest run --reporter=verbose` | `math.test.ts > math > rejects division by zero` | 1.0, `resolved=True` |

## Test plan

- [x] `uv run --all-extras pytest -q`: 1957 passed, 6 skipped (live-network tests)
- [x] `uv run ruff check .` and `uv run ruff format --check .`
- [x] New `tests/test_vitest_output.py` runs every case against both parsers. Its fixtures are verbatim runner output with local paths shortened to `/repo`: vitest 5.0.1 before and after the fix, the default reporter, two projects with and without color, vitest 3.2.7, and colored jest 30.5.0.
- [x] 21 of the 23 new tests fail on main (including the `normalize_test_cmds_for_runtime` case), and all pass with this change.
- [x] The standalone verifier, run with `python -I`, auto-detects the runner from `npx vitest run --reporter=verbose` and grades the agent patch at reward 1.0.
- [x] Pathological 200k-character lines (repeated `(retry x1)`, ` [`, digits, ` > `, escape fragments) parse in under 100 ms.
- [x] Manually ran the normalized command on vitest 3.2.7, 4.1.11 and 5.0.1.

## Out of scope

- Wrappers such as `npm test` get no reporter flag, as before (flags would need to go after `--`), so a vitest suite behind `npm test` still uses the default reporter.
- Mocha's spec reporter (`✔`, `1) title`) still doesn't parse. I'll file that separately.
- Vitest 2's tree-style verbose output, project names containing spaces with color, and logs that mix jest and vitest runs.
- Tasks generated before this change keep their baked verifier.

Closes #142


adithya-s-k:
Thanks for the thorough reproduction. Verified colored Vitest output, stable test names and standalone verifier grading. Tests and CI pass; merged.

PR #45, opened 2026-05-26
pr_runtime: graded F2P/P2P reward + 100-env dataset (Arc 2)

adithya-s-k:
Arc 2 of the v0.8.3 pipeline-optimization effort. Closes #44.

## Summary
Upgrades `pr_runtime` from a **binary** test-execution reward to a **graded** F2P/P2P reward that gives RL training a dense gradient while still serving strict SWE-bench eval — plus four correctness fixes found by auditing the output before scaling. Ships **99 oracle-verified environments**.

## Graded reward
- New `_pr_runtime_verifier.py` (pure stdlib, base64-baked): `reward = f2p_rate × p2p_rate` → `reward.txt` (dense training signal), strict `resolved` bool + breakdown → `reward.json` (eval signal). Ports the 4 log parsers (pytest/go/cargo/jest); exit-code fallback on unparseable output.
- **Oracle invariant**: gold patch flips all F2P + keeps all P2P → reward 1.0 (oracle gate unchanged).
- Difficulty + coverage metadata per task (`f2p_count`, `p2p_count`, `loc_changed`, `difficulty`).

## Scale audit — four real fixes
1. **F2P detection silently broken on real repos**: validation markers went to stderr (`set -x`) while test output went to stdout, + 20 KB truncation vs ~140 KB suites → **0 tests parsed → every candidate dropped**. Fix: stdout markers + parse full output. (`pallets/click`: emission 0 → 12.)
2. **Instructions leaked the solution**: built from the PR body (commit SHAs, fix approach, grading-test names — agent exploited a leaked SHA 28×). Fix: source from the linked **issue**, extend leak-strip, drop "Tests added" sections, filter non-bug PRs (backport/cherry-pick/release/revert).
3. **`git clean -fdx` wiped polyglot deps** (node_modules/target/vendor) → Node/Rust yielded 0. Fix: exclude dep dirs.
4. **Missing ca-certificates** → HTTPS git fetch failed on alpine/Node images. Fix: install ca-certificates.

Also: **multi-image datasets now publish in inline mode** (a multi-repo runtime dataset has one image per repo; each task bakes its own recipe — no registry creds needed).

## Reference dataset
**https://huggingface.co/datasets/AdithyaSK/repo2rlenv-pr-runtime** — 100 oracle-verified envs (**63 Python / 37 Go**, 13 repos), in the collection. Difficulty: trivial 18 / small 31 / medium 38 / large 13.

## Validation
[...]

PR #145, opened 2026-09-18
pr_diff: ship oracle + verifier via tests/, not baked in image

### LINKED ISSUE #144: pr_diff bakes the oracle into the agent's image, so the agent can read and apply it for a perfect reward
### Description

Every `pr_diff` task's `environment/Dockerfile` base64-bakes the oracle patch, the instruction, and the verifier into `/verifier/` in the image (`pr_diff.py`, `build_pr_diff_environment_dockerfile`):

```
RUN echo "<b64>" | base64 -d > /verifier/oracle.patch
RUN echo "<b64>" | base64 -d > /verifier/instruction.md
RUN echo "<b64>" | base64 -d > /verifier/verifier.py
```

That is the same container the agent gets a shell in, and it runs as root. So the agent can read `/verifier/oracle.patch` — the reference fix — and apply it. `tests/test.sh` then scores the agent's diff against that same oracle, yielding `reward = 1.0` on every deterministic component without the agent solving anything. For an RL training environment this is a reward-hack a policy under optimization pressure will find.

The runtime pipelines already avoid this: `pr_runtime` ships its grading assets as plain `tests/` files (`_runtime_aux_files`), and Harbor uploads a task's `tests/` into the container **only at verification time**, never during the agent phase. `pr_diff` predates that pattern — the baking landed in #40, the plain-`tests/` pattern in #45 — and was never migrated. The same class of leak was fixed for `code_instruct` / `equivalence_tests` in #55.

### Reproduction

The Dockerfile builder bakes the oracle directly, so no network or dataset is needed:

```python
from repo2rlenv.pipelines.pr_diff import build_pr_diff_environment_dockerfile
df = build_pr_diff_environment_dockerfile(
    repo_url="https://github.com/pallets/click",
    base_commit="<sha>",
    oracle_diff="diff --git a/x b/x\n+the_fix\n",
    instruction="...",
)
assert "base64 -d > /verifier/oracle.patch" in df   # the oracle is in the agent's image
```

Building any emitted task's `environment/` and shelling into it shows the same — the agent's container has `/verifier/oracle.patch`, and `git apply /verifier/oracle.patch` followed by the baked `tests/test.sh` scores `reward.txt = 1.000000`, all six components 1.0. (This also holds for any published `pr_diff` task, e.g. `AdithyaSK/repo2rlenv-pr-diff`.)

### Expected behavior

The oracle must not be reachable from the agent's environment. Ship the oracle, verifier, and instruction as `tests/` aux files (the `pr_runtime` pattern) so Harbor delivers them only at verify time; the agent's image contains the repo at `base_commit` and nothing else.

### Impact

Every `pr_diff` task generated to date is affected, including the published 181-task reference dataset. Existing datasets keep the verifier they were emitted with, so they need re-emission to pick up the fix; the generation-side change makes future tasks safe.

### Residual (separate)

Even with the oracle out of the image, the verifier still runs in the agent's own container (shared mode), so `PYTHONPATH` / `sitecustomize` / shadowed-binary tampering is a distinct, deeper vector. Harbor's `[verifier] environment_mode = "separate"` is the fix for that and is a larger change; I'd keep it to a follow-up.

I have a fix with tests, verified end to end through Harbor; PR to follow.

surajsharan:
## Summary

- **Close the reward-hack.** `pr_diff` base64-baked the oracle patch, instruction, and verifier into `/verifier/` in the agent's own image, so an agent could `git apply /verifier/oracle.patch` and score a perfect reward without solving anything. They now ship as `tests/{oracle.patch, verifier.py, instruction.md}` aux files, which Harbor uploads only at verify time — the exact plain-`tests/` pattern `pr_runtime` has used since #45, and the same class of fix as #55. The agent's image now contains the repo at `base_commit` and nothing else.
- **Second, named change: the reward file is cleared before the verifier runs.** `pr_diff`'s `test.sh` had no guard, so an agent that broke `python3` and pre-wrote `/logs/verifier/reward.txt` kept its value. `test.sh` now `rm -f`s the reward files first, mirroring Harbor's own separate-verifier `empty_dirs([verifier_dir])`. The verifier is the sole authority on the score.
- `validate --deep` now checks a runnable `pr_diff` task ships `tests/{verifier.py, oracle.patch, instruction.md}` (parity with the `pr_runtime` graded-verifier check), so a task that would build but score nothing is caught statically.
- Docs corrected: the layout blocks in `pipelines/README.md` and `pr_diff.md`, and RFC 0001, described the baked `/verifier/` layout that this removes.

## Test plan

- [x] `uv run --all-extras pytest -q` — 1933 passed, 10 skipped (new/updated tests in `test_pipeline_pr_diff.py`, `test_validation.py`, `test_pr_diff_clone.py`)
- [x] `ruff check .` + `ruff format --check .` clean; `mkdocs build` clean (0 warnings)
- [x] `repo2rlenv validate --deep` passes on a freshly emitted task and flags one with the `tests/` assets removed
- [x] **Leak reproduced on `main`, absent after** (Docker Desktop / WSL2). On a task emitted with this branch's builders, the built image has no `/verifier`, no `/tests`, and no `oracle.patch` anywhere — `find / -name oracle.patch` is empty — so the `git apply /verifier/oracle.patch` exploit is impossible.
- [x] **End to end still scores 1.0.** `harbor run -a oracle -e docker` on the fixed task → `reward.txt = 1.000000`, all deterministic components 1.0, `judge_status: no_api_key` (renormalized). The verifier runs correctly from the `tests/` upload delivered at verify time.

## Out of scope

- **Existing published datasets.** The change is generation-side; already-emitted `pr_diff` tasks (including `AdithyaSK/repo2rlenv-pr-diff`) keep their baked verifier and need re-emission to pick up the fix. A migration/re-bake tool could be a follow-up — and, since the verifier is now a plain `tests/verifier.py` file, upgrading it in place becomes a file swap rather than an image rebuild.
- **The shared-verifier residual.** The verifier still runs in the agent's container, so `PYTHONPATH` / `sitecustomize` / shadowed-binary tampering remains. The fix is Harbor's `[verifier] environment_mode = "separate"`, a larger change worth its own discussion.
- Release-notes entry, left for the release cut.

Closes #144


adithya-s-k:
Thanks, this closes the baked-oracle leak for new exports. I added a stale JSON reward cleanup, executable no-op/oracle regressions, and migration guidance. All tests and CI pass; merged. Existing published tasks still need repair, tracked in #155.

